PR #24, opened 2020-01-31
add wal-g

patsevanton:


vitabaks:
@patsevanton thanks for your PR.

You tested of create replica from a backup using wal-g?

patsevanton:
No, i don`t test patroni + wal-g.

vitabaks:
@patsevanton 

I suggest moving the variable "wal_g_ver" below the variable "patroni_create_replica_methods".

And the comment "# version for deploy etcd cluster" needs to be fixed.

It is also worth adding the variable "wal_g_install" which will be enabled by default, but if necessary, you can disable the wal-g installation.
There may be a situation where you do not need to install wal-g using this playbook, for example, when the wal-g is already installed in any other method.

Example:
```
# WAL-G
wal_g_install: "true" # or "false"
wal_g_install_version: "v0.2.14" # version for install wal-g (if patroni_create_replica_methods.wal_g is defined and wal_g_Install: "true")

```

PR #28, opened 2020-02-10
add install wal-g

patsevanton:


patsevanton:
Конфликтов нет. Проверю попозже.

vitabaks:
@patsevanton 

I suggest moving the variable "wal_g_ver" below the variable "patroni_create_replica_methods".

And the comment "# version for deploy etcd cluster" needs to be fixed.

It is also worth adding the variable "wal_g_install" which will be enabled by default, but if necessary, you can disable the wal-g installation.
There may be a situation where you do not need to install wal-g using this playbook, for example, when the wal-g is already installed in any other method.

Example:
```
# WAL-G
wal_g_install: "true" # or "false"
wal_g_install_version: "v0.2.14" # (if patroni_create_replica_methods.wal_g is defined and wal_g_install: "true")

```

vitabaks:
The wal-g.yml file does not contain the task of preparing the wal-g configuration.

patsevanton:
image

vitabaks:
`when: wal_g_install == "true" and 'wal_g' in patroni_create_replica_methods`
or:
`when: wal_g_install == "true" and patroni_create_replica_methods.wal_g is defined`

patsevanton:
https://github.com/patsevanton/postgresql_cluster/commit/dcf8eaa8183d54e41fd86ea203dcf3cb773eeb35

This is simple comment

vitabaks:
d7bea1379b2588ea41644e699457ebf2a25972ce - this is superfluous, please remove.
These tasks are necessary only for debugging and should not fall into the PR.

patsevanton:
Wait release after https://github.com/wal-g/wal-g/pull/574

patsevanton:
Create new PR. Close

PR #62, opened 2020-09-21
Changed conditional pgbouncer.ini.j2

FabianHardt:
It's a wrong conditional here. "Not" should be removed for inserting VIP.


vitabaks:
It's not a mistake.

pgbouncer is available via cluster_vip only when there is no load balancer (with_haproxy_load_balancing: false).
When the load balancer is in use, there is no need to listen to cluster_vip.

FabianHardt:
I'm sorry, I just changed my configs while stepwise executing. That's why my haproxy.cfg didn't routed to pgbouncer.
Your code is just fine!

PR #72, opened 2020-11-17
Update main.yml

Donik-ops:


vitabaks:
the -q option has no benefit.
```
-q, --quiet           quiet operation
```
Why did you decide to add it?

But we need the -C option so as not to try to update the cache
```   
 -C, --cacheonly       run entirely from system cache, don't update cache
```
otherwise you may get an error like this:
```
Failed to download metadata for repo 'AppStream'
Error: Failed to download metadata for repo 'AppStream'
```

PR #74, opened 2020-12-01
Use update_cache in apt module.

rasooll:
Better is to update the cache when installing packages.

### DIFF
--- a/roles/packages/tasks/main.yml
+++ b/roles/packages/tasks/main.yml
@@ -58,6 +58,7 @@
   apt:
     name: "{{ item }}"
     state: latest
+    update_cache: yes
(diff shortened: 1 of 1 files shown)


vitabaks:
Hi @rasooll 

You don't have to do this for every packet in the loop.

We update the apt cache once during the launch of the playbook.
See the "Update apt cache" [task](https://github.com/vitabaks/postgresql_cluster/blob/master/deploy_pgcluster.yml#L42) in deploy_pgcluster.yml

rasooll:
Hi @vitabaks 

I saw "Update apt cache" and I know my changes aren't necessary, so I close my pull request.

Thank you for your description.

PR #77, opened 2020-12-11
Restrict standard local port range

mol-lux:
Using a local port range of 1024-65535 is risky because etcd listens on port 2379 and 2380, pgbouncer on port 6432, PostgreSQL on port 5432 and Patroni on 8008. This can lead to collision.

In my case, etcd established a connection to another cluster member on port 8008, which meant that Patroni wasn't able to start because port 8008 was already in use.

10000 stays clear of all predefined ports.


vitabaks:
@mol-lux Ok, thanks!

You're right, it's safer this way.
Considering that the main thing for us is the stability of the cluster.

Initial commit (for history):
https://github.com/vitabaks/postgresql_cluster/commit/6e0a6d9714cde75e9919aead6209ee06cc8a12a1 **Increase the range of local ports available for outgoing connections.**

PR #86, opened 2021-01-08
Make Patroni REST API port configurable

mol-lux:
In our setup, we would like to be able to configure the port that the Patroni API is listening on.

I did a find/replace on the codebase and replaced it with a new custom variable, ` patroni_restapi_port`, which is defined in `vars/main.yml`

Thanks again for this awesome role, I really like it!


vitabaks:
> In our setup, we would like to be able to configure the port that the Patroni API is listening on.

why?

vitabaks:
All checks have failed

The error appears to be in '/github/workspace/roles/patroni/tasks/main.yml': line 757, column 16, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

      wait_for:
        port: {{ patroni_restapi_port }}
               ^ here

Issue with missing quotes. Always quote template expression brackets in ansible tasks. 
For instance:

      wait_for:
        port: "{{ patroni_restapi_port }}"

PR #114, opened 2021-08-22
Enable wal_compression by default

NikolayS:
Why: https://gitlab.com/postgres-ai/postgresql-consulting/tests-and-benchmarks/-/issues/15


vitabaks:
You're reading my mind @NikolayS

Thanks!

PR #136, opened 2021-12-27
Added the possibility to add options to PostgreSQL roles.

sjstoelting:
I've extended the creation of PostgreSQL roles to support options like SUPERUSER, NOLOGIN, etc.


vitabaks:
Hi @sjstoelting 

Thanks for your PR!

PR #137, opened 2021-12-27
Changed PostgreSQL password encryption from md5 to scram-sha-256.

sjstoelting:
I switched password encryption away from insecure md5 to scram-sha-256, which is default in PostgreSQL 14.

### DIFF
--- a/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -15,11 +15,11 @@ listen_addr = {{ hostvars[inventory_hostname]['inventory_hostname'] }}
 {% endif %}
 listen_port = {{ pgbouncer_listen_port | default(6432) }}
 unix_socket_dir = /var/run/postgresql
-auth_type = md5
+auth_type = scram-sha-256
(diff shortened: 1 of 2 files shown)


vitabaks:
Note that older clients might lack support for the SCRAM authentication mechanism, and hence not work with passwords encrypted with SCRAM-SHA-256.

I think it's worth adding a comment:

"`  # or ''md5" if your clients do not work with passwords encrypted with SCRAM-SHA-256`"

sjstoelting:
It might also be necessary, to add a hint, that the PgBouncer configuration has to be changed, too.
Otherwise the login through PgBouncer will fail.

" # or ''md5" if your clients do not work with passwords encrypted with SCRAM-SHA-256, you have to change the PgBouncer configration in /roles/pgbouncer/templates/pgbouncer.ini.j2, too"

vitabaks:
>  you have to change the PgBouncer configuration in /roles/pgbouncer/templates/pgbouncer.ini.j2, too

I think it's better to add the `pgbouncer_auth_type` variable to vars/main

sjstoelting:
I'll create a new pull request, where I changed the authentication method handled by a variable.

PR #138, opened 2021-12-27
Scram passwords

sjstoelting:
I've added the variable password_encryption in vars/main.yml to have a single point for the encryption method.
The variable is used in PostgreSQL options, role examples, and PgBouncer options.

### DIFF
--- a/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -15,11 +15,11 @@ listen_addr = {{ hostvars[inventory_hostname]['inventory_hostname'] }}
 {% endif %}
 listen_port = {{ pgbouncer_listen_port | default(6432) }}
 unix_socket_dir = /var/run/postgresql
-auth_type = md5
+auth_type = scram-sha-256
 auth_file = /etc/pgbouncer/userlist.txt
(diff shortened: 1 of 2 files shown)


vitabaks:
Thanks @sjstoelting  it looks great!

Do you have tried using the `postgresql_parameters.password_encryption` variable for `pgbouncer_auth_type` instead of defining a new variable?

vitabaks:
> Do you have tried using the postgresql_parameters.password_encryption variable for pgbouncer_auth_type (and etc..) instead of defining a new variable?

We can leave it like this for now, and then I'll think about how to improve it

sjstoelting:
No, I haven't considered  postgresql_parameters.password_encryption, let me give it a try.

sjstoelting:
I've tried it with postgresql_parameters.password_encryption, that did not work.
Though, it could stay as is, or roles/pgbouncer/templates/pgbouncer.ini.j2 could use the variable  `password_encryption`.

vitabaks:
> I've tried it with postgresql_parameters.password_encryption, that did not work.
Though, it could stay as is, or roles/pgbouncer/templates/pgbouncer.ini.j2 could use the variable password_encryption.

Well, thank you for taking the time to do this.

I modify this variables later.

PR #139, opened 2022-01-08
Pgbackrest srv

sjstoelting:
I have added the possibility to have pgBackRest running on a dedicated server. This has two advantages, one can take backups from one of the followers instead of taking backups from the leader, and it moves the compression of the backup away from the database server.
The usage is optional configured in pgbackrest_srv_install. But it does require pgbackrest_install, too.

### DIFF
--- a/roles/patroni/templates/pg_hba.conf.j2
+++ b/roles/patroni/templates/pg_hba.conf.j2
@@ -96,3 +96,9 @@
 {% for host in groups['postgres_cluster'] %}
   host      replication              {{ patroni_replication_username }}               {{ hostvars[host]['inventory_hostname'] }}/32         {{ postgresql_password_encryption_algorithm }}
 {% endfor %}
+# For pgBackRest from external server
+{% if pgbackrest_srv_install -%}
+{% for pgbackrestsrv in groups['pgbackrest'] %}
+  host      all              postgres               {{ hostvars[pgbackrestsrv]['inventory_hostname'] }}/32                md5
+{% endfor %}
+{% endif %}
--- a/roles/pgbackrest_srv/tasks/main.yml
+++ b/roles/pgbackrest_srv/tasks/main.yml
@@ -1,0 +1,115 @@
+---
+# yamllint disable rule:line-length
+
+- block:  # Debian pgdg repo
+    - name: Make sure the gnupg and apt-transport-https packages are present
+      apt:
+        pkg:
+          - gnupg
+          - apt-transport-https
+        state: present
+
+    - name: Make sure pgdg apt key is installed
+      apt_key:
+        id: ACCC4CF8
+        url: https://apt.postgresql.org/pub/repos/apt/ACCC4CF8.asc
+
+    - name: Make sure pgdg repository is installed
+      apt_repository:
+        repo: "deb https://apt.postgresql.org/pub/repos/apt/ {{ ansible_distribution_release }}-pgdg main"
+
+    - name: Update apt cache
+      apt:
+        update_cache: true
+  environment: "{{ proxy_env | default({}) }}"
+  when:
+    - installation_method == "repo"
+    - ansible_os_family == "Debian"
+    - pgbackrest_install_from_pgdg_repo|bool
+  tags: pgbackrest, pgbackrest_repo, pgbackrest_install, pgbackrest_srv_install
+
+- block:  # RedHat pgdg repo
+    - name: Get pgdg-redhat-repo-latest.noarch.rpm
+      get_url:
+        url: "https://download.postgresql.org/pub/repos/yum/reporpms/EL-{{ ansible_distribution_major_version }}-x86_64/pgdg-redhat-repo-latest.noarch.rpm"  # noqa 204
+        dest: /tmp/
+        timeout: 30
+        validate_certs: false
+
+    - name: Make sure pgdg repository is installed
+      package:
+        name: /tmp/pgdg-redhat-repo-latest.noarch.rpm
+        state: present
+
+    - name: Clean yum cache
+      command: yum clean all
+      args:
+        warn: false
+      when: ansible_distribution_major_version == '7'
+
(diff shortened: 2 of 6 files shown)


vitabaks:
I don't understand yet why the current pgbackrest role is not enough, the current version of the playbook installs pgbackrest and prepares the configuration for restore from a dedicated pgbackrest server via ssh or use s3 (and other storages)

vitabaks:
Do you want to transfer the task of configuring the backup server to this playbook?

Then it is necessary to automate and configure the cron job for Full and Incr backups.

Also, I think it should be one pgbackrest role and not two roles with the same code

roles/patroni/templates/pg_hba.conf.j2
```diff
@@ -96,3 +96,9 @@
 {% for host in groups['postgres_cluster'] %}
   host      replication              {{ patroni_replication_username }}               {{ hostvars[host]['inventory_hostname'] }}/32         {{ postgresql_password_encryption_algorithm }}
 {% endfor %}
+# For pgBackRest from external server
+{% if pgbackrest_srv_install -%}
+{% for pgbackrestsrv in groups['pgbackrest'] %}
+  host      all              postgres               {{ hostvars[pgbackrestsrv]['inventory_hostname'] }}/32                md5
```
vitabaks, line 102: as far as I know, the dedicated backup server works over ssh and there is no need to additionally add it to pg_hba on postgresql servers.

roles/pgbackrest_srv/tasks/main.yml
```diff
@@ -0,0 +1,115 @@
+---
```
vitabaks, line 1: This code duplicates roles/pgbackrest/tasks/main.yml

I think it should be one pgbackrest role and not two roles with the same code

roles/pgbackrest_srv/tasks/ssh_keys.yml
```diff
@@ -0,0 +1,80 @@
+---
```
vitabaks, line 1: This code duplicates roles/pgbackrest/tasks/ssh_keys.yml

I think it should be one pgbackrest role and not two roles with the same code

roles/pgbackrest_srv/tasks/system.yml
```diff
@@ -0,0 +1,159 @@
+---
```
vitabaks, line 1: `system.yml` these are variables 

perhaps they are here by mistake

roles/pgbackrest_srv/templates/pgbackrest.conf.j2
```diff
@@ -0,0 +1,13 @@
+[global]
```
vitabaks, line 1: please use roles/pgbackrest

it can be a separate template, for example `pgbackrest.server.conf.j2`

vars/main.yml
```diff
@@ -319,6 +319,7 @@ wal_g_patroni_cluster_bootstrap_command: "wal-g backup-fetch {{ postgresql_data_
 
 # pgBackRest
 pgbackrest_install: false  # or 'true'
+pgbackrest_srv_install: false  # or 'true'
```
vitabaks, line 322: I think the `pgbackrest_install` variable should be enough

vars/main.yml
```diff
@@ -331,6 +332,7 @@ pgbackrest_conf:
     - {option: "log-level-file", value: "detail"}
     - {option: "log-path", value: "/var/log/pgbackrest"}
     - {option: "repo1-type", value: "{{ pgbackrest_repo_type |lower }}"}
+#    - {option: "repo1-retention-full", value: 5} #  Take care, not to fill up your disks with backups
```
vitabaks, line 335: I think it's worth uncommenting out this parameter, but specify the default value of 1 or 2

> Take care, not to fill up your disks with backups

this comment seems superfluous to me

vars/main.yml
```diff
@@ -339,6 +341,20 @@ pgbackrest_conf:
     - {option: "process-max", value: "2"}
     - {option: "recovery-option", value: "recovery_target_action=promote"}
 #    - {option: "", value: ""}
+
+pgbackrest_srv_conf:
+  global:  # [global] section
+    - {option: "log-level-file", value: "detail"}
+    - {option: "log-path", value: "/var/log/pgbackrest"}
+    - {option: "repo1-type", value: "{{ pgbackrest_repo_type |lower }}"}
+#   - {option: "repo1-retention-full", value: 5} #  Take care, not to fill up your disks with backups
```
vitabaks, line 350: I think it's worth uncommenting out this parameter, but specify the default value of 1 or 2

> Take care, not to fill up your disks with backups

this comment seems superfluous to me

vars/main.yml
```diff
@@ -339,6 +341,20 @@ pgbackrest_conf:
...
+    - {option: "repo1-path", value: "/var/lib/pgbackrest"}
+    - {option: "repo1-host-user", value: "{{ pgbackrest_repo_user }}"}
+  stanza:
+    - {option: "process-max", value: "2"}
+    - {option: "recovery-option", value: "recovery_target_action=promote"}
+    - {option: "backup-standby", value: "y"}  # Don't take backups from a primary, move the work to a secondary, especially because of compression
```
vitabaks, line 356: I suggest replacing the comment with the one specified in the pgbackrest documentation

`# Enable backup from standby to reduce load on the primary cluster. `

vars/main.yml
```diff
@@ -339,6 +341,20 @@ pgbackrest_conf:
     - {option: "process-max", value: "2"}
     - {option: "recovery-option", value: "recovery_target_action=promote"}
 #    - {option: "", value: ""}
+
+pgbackrest_srv_conf:
```
vitabaks, line 345: I think a detailed comment is needed here for what the pgbackrest_srv_conf variables are for, since they are very similar to the pgbackrest_conf variables

PR #187, opened 2022-07-22
Adding PostgreSQL schemas

chlordk:
I need to 'CREATE SCHEMA' in PostgreSQL. It is a small task which could just be added after deployment but then it has to be written in my documentation. So here is an Ansible play to do the task if it useful to anyone.

### DIFF
--- a/deploy_pgcluster.yml
+++ b/deploy_pgcluster.yml
@@ -185,6 +185,9 @@
     - role: postgresql-databases
       when: is_master == "true" and postgresql_databases | length > 0
 
+    - role: postgresql-schemas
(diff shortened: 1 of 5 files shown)


vitabaks:
Thanks @chlordk !

chlordk:
@vitabaks Your welcome.

I was looking around to see which parameters in vars/main.yml you are using but could not find them.
A good test of schema could be:

`    postgresql_extension:`
`      - {ext: "pg_stat_statements", db: ""mydatabase", schema: "myschema"}`

If this works mostly eveything about schemas has been tested.

PR #188, opened 2022-07-26
Add remove_cluster.yml

chlordk:
An Ansible script for easy undeployment of a PostgreSQL cluster while experimenting with different configurations.

Could also be called 'undeploy_cluster.yml'.

To prevent unintentionally undeployment 'remove_postgres' and 'remove_etcd' should be set to 'true'.

TODO: If it is possible to reuse the variable 'pgbackrest_conf global repo1-path' that will come in handy. Otherwise a new variable 'pgbackrest_repo1-path' could be made and that then used for the config option also.

### DIFF
--- a/remove_cluster.yml
+++ b/remove_cluster.yml
@@ -1,0 +1,44 @@
+---
+
+- hosts: postgres_cluster
+  vars:
+    remove_postgres: false  # or 'true' to remove the database
+  become: true
+  vars_files:
+    - vars/main.yml
+    - vars/{{ ansible_os_family }}.yml
+  tasks:
+    - name: Stop Patroni service
+      service:
+        name: patroni
+        state: stopped
+    - name: Delete PostgreSQL database content
+      file:
+        path: "{{ postgresql_data_dir }}"
+        state: absent
+      when: remove_postgres|bool
+    - name: Delete PgBackRest repository
+      file:
+        # path: pgbackrest_conf global repo1-path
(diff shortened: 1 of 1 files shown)


vitabaks:
Could you also update the README?

https://github.com/vitabaks/postgresql_cluster#how-to-start-from-scratch

chlordk:
> it would be nice if such a playbook completely removed all packages

I see some problems with that:

1. If you remove say 'gcc' and that should be used for something else. If you only use this script for testing then it is not a problem.
2. The list of packages: Is it possible to have only **one** list which can be used for install and remove?
3. Is it easy to remove tar-balls? Will *make uninstall* do it?

Safety pin: Is there an easy way to make sure that the script wont work before you have edited it? This should prevent someone run in a production environment.

Here is an instruction.

**How to start from scratch**
Should you need to start from very beginning, use the playbook `remove_cluster.yml`.

To prevent the script to be used by accident in a production environment, edit `remove_cluster.yml` and remove the *safety pin*. Change these variables:

- remove_postgres: true
- remove_etcd: true

Run the script and all the data are gone. A new installation can now be made from scratch.

:heavy_exclamation_mark: Be careful not to copy this script without *safety pin* to the production environment.

vitabaks:
ok, for now, let there be only the removal of a minimum number of packages, and data for the purpose of redeploying.

Thanks @chlordk

PR #198, opened 2022-09-01
Make image path relative

chlordk:
A clone should refer to own images,
not the master.


chlordk:
These errors has nothing to do with the patch I send.
Is there an error in the test procedure?

vitabaks:
Thanks @chlordk

PR #203, opened 2022-09-08
Set vm.nr_hugepages playbook 

chlordk:


chlordk:
I'm not sure why this came with the PR???
> Merge branch 'vitabaks:master' into master

It was not supposed to be there.
Should I close this and make a new PR?

chlordk:
@vitabaks Do you want a description of how to calculate `huge_pages` in README.md?
It will based on this description:
[Why Linux HugePages are Super Important for Database Servers: A Case with PostgreSQL](https://www.percona.com/blog/why-linux-hugepages-are-super-important-for-database-servers-a-case-with-postgresql/)

vitabaks:
Thanks @chlordk 

I think it should be a role and not a poeybook.

chlordk:
> I think it should be a role and not a plaeybook.

Oh, I just see that it is already there. I Just specified `vm.nr_hugepages` in `vars/system.yml` and it works fine.

So please reject this PR.

PR #206, opened 2022-09-23
Check cluster already installed

chlordk:
Make deployment stop if the cluster is already installed.

### DIFF
--- a/deploy_pgcluster.yml
+++ b/deploy_pgcluster.yml
@@ -19,6 +19,20 @@
   gather_facts: true
   tags: always
 
+  vars_files:
+      - vars/main.yml
+      - "vars/{{ ansible_os_family }}.yml"
+  tasks:
+    - name: Check cluster already installed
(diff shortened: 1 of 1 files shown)


vitabaks:
@chlordk 

We already have a similar check that does not allow deleting data for an already existing PostgreSQL instance

`check that data directory "{{ postgresql_data_dir }}" is not initialized`

https://github.com/vitabaks/postgresql_cluster/blob/master/roles/patroni/tasks/main.yml#L380

vitabaks:
I understand you just want to perform this check at an earlier stage

vitabaks:
I prefer not to add these tasks to the "Gathering facts from all servers" block \
for pre-checks there are pre_tasks for hosts: postgres_cluster

You can add "Update apt cache" before the task

chlordk:
> We already have a similar check

Sorry. I must have missed it.

Just reject this PR.

PR #226, opened 2022-12-04
add the ability to install new versions of WAL-G

vitabaks:
wal_g_version: "2.0.1"


vitabaks:
Problem when building from source code for the Red Hat family

Issue: https://github.com/wal-g/wal-g/issues/1412

PR #227, opened 2022-12-06
 delete node or balancer 

prensgold:
we want to scale-down operations with delete node or balancer when the node or balancer gives an error that cannot be fixed.
can you this feature add your playbook ?


vitabaks:
@prensgold a lot of changes and including conflicts \
please update your branch and recreate the PR

PR #231, opened 2022-12-14
Changes in confd and haproxy roles.

artemsafiyulin:
1. If generate confd.toml nodes without "http://", then service dont start with error:
FATAL Near line 4 (last key parsed 'nodes'): Invalid float value: "10.128.64.140"

2. A variable "postgresql_cluster_maintenance" has been added to config_cluster.yml to not generate the haproxy.cfg file from being generated by the haproxy task when confd is used. (deploy_pgcluster.yml - no changes)

### DIFF
--- a/roles/haproxy/tasks/main.yml
+++ b/roles/haproxy/tasks/main.yml
@@ -354,7 +354,7 @@
     owner: haproxy
     group: haproxy
   notify: "restart haproxy"
-  when: add_balancer is not defined or not add_balancer|bool
+  when: (add_balancer is not defined or not add_balancer|bool) and (postgresql_cluster_maintenance is not defined or not postgresql_cluster_maintenance|bool)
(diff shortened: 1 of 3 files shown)


roles/haproxy/tasks/main.yml
```diff
@@ -354,7 +354,7 @@
     owner: haproxy
     group: haproxy
   notify: "restart haproxy"
-  when: add_balancer is not defined or not add_balancer|bool
+  when: (add_balancer is not defined or not add_balancer|bool) and (postgresql_cluster_maintenance is not defined or not postgresql_cluster_maintenance|bool)
```
vitabaks, line 357:
```suggestion
  when: (add_balancer is not defined or not add_balancer|bool) and
        (postgresql_cluster_maintenance is not defined or not postgresql_cluster_maintenance|bool)
```

PR #242, opened 2023-01-19
Fix /etc/environment error

anyuta1166:
I've noticed the following error in system log:

`systemd[3456620]: /usr/lib/environment.d/99-environment.conf:2: invalid variable name "export PATRONICTL_CONFIG_FILE", ignoring.`

/usr/lib/environment.d/99-environment.conf is a symlink to /etc/environment.

Documentation says that the /etc/environment file must consist of simple NAME=VALUE pairs on separate lines.

According to this we should probably remove "export" before the variable name.

### DIFF
--- a/roles/patroni/tasks/main.yml
+++ b/roles/patroni/tasks/main.yml
@@ -980,8 +980,8 @@
   lineinfile:
     dest: "/etc/environment"
     state: present
-    regexp: "^export PATRONICTL_CONFIG_FILE"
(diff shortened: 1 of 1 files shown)


vitabaks:
@anyuta1166 That makes sense, thanks!

> Ensure that the information in the environment file is in the Name=Value format. Unlike profile scripts, the environment file is not a shell script and does not accept data in any format other than the Name=Value format.

https://www.ibm.com/docs/en/aix/7.2?topic=files-environment-file

PR #247, opened 2023-01-30
Add Patroni raft

tanganellilore:
Hi @vitabaks,

I added support to Patrini raft (present here from patroni v2). This allow us to avoid the usage of etcd and consul.

Moreover  I added:

- a script for HA in keepalived to discover the master and assign correctly keepalived
- patroni_watchdog configuration (and respective patroni_watchdog_device)
- added doc side for raft

I only missed the architettural image for patroni raft, because i don't find the original code of image (i think draw.io).

Let me know if I need to change something.
Thanks for your work!

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,36 +22,48 @@ In addition to deploying new clusters, this playbook also support the deployment
 
 
 ## Index
-- [Cluster types](#cluster-types)
-    - [[Type A] PostgreSQL High-Availability with HAProxy Load Balancing](#type-a-postgresql-high-availability-with-haproxy-load-balancing)
-    - [[Type B] PostgreSQL High-Availability only](#type-b-postgresql-high-availability-only)
-    - [[Type C] PostgreSQL High-Availability with Consul Service Discovery (DNS)](#type-c-postgresql-high-availability-with-consul-service-discovery-dns)
-- [Compatibility](#compatibility)
-    - [Supported Linux Distributions:](#supported-linux-distributions)
-    - [PostgreSQL versions:](#postgresql-versions)
-    - [Ansible version](#ansible-version)
-- [Requirements](#requirements)
-- [Port requirements](#port-requirements)
-- [Recommendations](#recommenations)
-- [Deployment: quick start](#deployment-quick-start)
-- [Variables](#variables)
-- [Cluster Scaling](#cluster-scaling)
-    - [Preparation:](#preparation)
-    - [Steps to add a new node:](#steps-to-add-a-new-node)
-    - [Steps to add a new banlancer node:](#steps-to-add-a-new-banlancer-node)
-- [Restore and Cloning](#restore-and-cloning)
-    - [Create cluster with pgBackRest:](#create-cluster-with-pgbackrest)
-    - [Create cluster with WAL-G:](#create-cluster-with-wal-g)
-    - [Point-In-Time-Recovery:](#point-in-time-recovery)
-- [Maintenance](#maintenance)
-- [Disaster Recovery](#disaster-recovery)
-    - [etcd](#etcd)
-    - [PostgreSQL (databases)](#postgresql-databases)
-- [How to start from scratch](#how-to-start-from-scratch)
-- [License](#license)
-- [Author](#author)
-- [Sponsor this project](#sponsor-this-project)
-- [Feedback, bug-reports, requests, ...](#feedback-bug-reports-requests-)
+- [PostgreSQL High-Availability Cluster :elephant: :sparkling\_heart:](#postgresql-high-availability-cluster-elephant-sparkling_heart)
+    - [Deploy a Production Ready PostgreSQL High-Availability Cluster (based on "Patroni" and DCS "etcd" or "consul"). Automating with Ansible.](#deploy-a-production-ready-postgresql-high-availability-cluster-based-on-patroni-and-dcs-etcd-or-consul-automating-with-ansible)
+  - [Index](#index)
+  - [Cluster types](#cluster-types)
+    - [\[Type A\] PostgreSQL High-Availability with HAProxy Load Balancing](#type-a-postgresql-high-availability-with-haproxy-load-balancing)
+          - [if variable "synchronous\_mode" is 'true' (vars/main.yml):](#if-variable-synchronous_mode-is-true-varsmainyml)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -26,6 +26,7 @@ In addition to deploying new clusters, this playbook also support the deployment
     - [[Type A] PostgreSQL High-Availability with HAProxy Load Balancing](#type-a-postgresql-high-availability-with-haproxy-load-balancing)
     - [[Type B] PostgreSQL High-Availability only](#type-b-postgresql-high-availability-only)
     - [[Type C] PostgreSQL High-Availability with Consul Service Discovery (DNS)](#type-c-postgresql-high-availability-with-consul-service-discovery-dns)
+    - [[Type D] PostgreSQL High-Availability with HAProxy Load Balancing and Patroni raft](#type-d-postgresql-high-availability-with-haproxy-load-balancing-and-patroni-raft)
```
vitabaks, line 29: I don't think dcs_type: "patroni_raft" deserves to be described as a separate scheme.

These are all the same type A and type B schemes, just instead of etcd, you can choose  "patroni_raft" in dcs_type

README.md
```diff
@@ -121,6 +122,41 @@ Example: `replica.postgres-cluster.service.dc1.consul`, `replica.postgres-cluste
 
 It requires the installation of a consul in client mode on each application server for service DNS resolution (or use [forward DNS](https://developer.hashicorp.com/consul/tutorials/networking/dns-forwarding?utm_source=docs) to the remote consul server instead of installing a local consul client).
 
+
+### [Type D] PostgreSQL High-Availability with HAProxy Load Balancing and Patroni raft
```
vitabaks, line 126: These are all the same type A and type B schemes, just instead of etcd, you can choose  "patroni_raft" in dcs_type

config_pgcluster.yml
```diff
@@ -69,6 +69,8 @@
     - name: Make sure the gnupg and apt-transport-https packages are present
       apt:
         pkg:
+          - curl
```
vitabaks, line 72: > Make sure the gnupg and apt-transport-https packages are present

the curl package is already defined in vars/debian - https://github.com/vitabaks/postgresql_cluster/blob/master/vars/Debian.yml#L43

roles/keepalived/templates/haproxy_check.sh.j2
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+CURRRENT_ROLE=$(curl http://{{ hostvars[host]['inventory_hostname'] }}:8008/master --silent | jq ' .role' | sed 's/"//g')
```
vitabaks, line 3: Why do you think that VIP should be run on a server with the role of master (leader)?

For the Type A scheme, it does not matter which of the servers is responsible for balancing

In the README you can find the following description for keepalived:

> In our configuration keepalived checks the status of the HAProxy service and in case of a failure delegates the VIP to another server in the cluster.

This is important to ensure the high availability of HAProxy

/bin/kill -0 `cat /run/haproxy/haproxy.pid`

I suggest not to make changes to this logic.

roles/patroni/templates/patroni.yml.j2
```diff
@@ -44,6 +44,13 @@ consul:
   checks: []
 {% endif %}
 
+{% if dcs_type == 'patroni_raft' %}
+raft:
+  data_dir: "{{ patroni_raft_data_dir }}"
+  self_addr: "{{ hostvars[inventory_hostname]['inventory_hostname'] }}:5010"
```
vitabaks, line 50: Please consider adding the `patroni_raft_port` variable

vars/main.yml
```diff
@@ -49,7 +49,7 @@ vip_manager_mask: "24"  # netmask for the virtual ip
 
 # DCS (Distributed Consensus Store)
 dcs_exists: false  # or 'true' if you don't want to deploy a new etcd cluster
-dcs_type: "etcd"  # or 'consul'
+dcs_type: "patroni_raft"  # or 'consul' or 'etcd'
```
vitabaks, line 52:
```suggestion
dcs_type: "etcd"  # or 'consul' or 'patroni_raft'
```

patroni_raft is currently in the beta stage. 

I suggest leaving the default configuration using etcd

besides:

https://github.com/zalando/patroni/blob/master/docs/releases.rst#version-300
> The RAFT support is deprecated. We will do our best to maintain it, but take neither guarantee nor responsibility for possible issues.

roles/keepalived/templates/haproxy_check.sh.j2
vitabaks: I think it makes sense for dcs_type == 'patroni_raft' 

Perhaps we could make a condition and add this check (leader role) only for patroni_raft

README.md
vitabaks: consider just adding a comment to the type A schema

README.md
vitabaks: consider just adding a comment about dcs_type: "patroni_raft" to the type A schema

roles/keepalived/templates/haproxy_check.sh.j2
vitabaks: > sed 's/"//g'

try `jq -r` instead of using sed

roles/patroni/tasks/main.yml
```diff
@@ -263,6 +263,16 @@
     mode: 0750
   when: patroni_log_destination == 'logfile'
   tags: patroni, patroni_conf
+
```
vitabaks, line 266:
```suggestion
 
```

Fix yamllint error: 266:1 [trailing-spaces] trailing spaces

vars/main.yml
vitabaks: apply before merging (after tests)

vitabaks:
Tests for RHEL do not seem to pass, they hang after starting replicas.

tanganellilore:
@vitabaks due to the fact that in v3 of patroni (as you also shared), team state that no more officially support internal RAFT.
So we can close this PR and probably you can state this on readme side

vitabaks:
@tanganellilore Ok. But you can open PR for the watchdog. I think anyone may need it.

tanganellilore:
I will do tomorrow morning with a specific PR.
Thanks!

PR #269, opened 2023-03-02
 Ability to install packages for patroni as a list

mrsrvman:
Example for installing patroni 3.0.1
```
patroni_packages:
  - patroni
  - patroni-etcd
  - python3-urllib3
```

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -40,6 +40,7 @@ system_packages:
   - python3-psycopg2
   - python3-setuptools
   - python3-pip
+  - python3-urllib3
   - curl
   - less
   - sudo
@@ -94,6 +95,8 @@ installation_method: "repo"  # (default)"repo" or "file"
 # Patroni package will be installed from the pip (by default).
 # You also have the option of choosing an patroni installation method using the deb package.
 patroni_installation_method: "pip"  # (default)"pip" or "deb"
+patroni_packages:
+  - patroni
 
 # (if patroni_installation_type: "pip")
 # Packages from your repository will be used to install instead of the pip repository.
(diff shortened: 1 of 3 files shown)


vitabaks:
@mrsrvman Please add the `patroni_packages` variable to `vars/Debian.yml` and `vars/RedHat.yml`

also add an example package name for consul

vars/Debian.yml
```diff
@@ -40,6 +40,7 @@ system_packages:
   - python3-psycopg2
   - python3-setuptools
   - python3-pip
+  - python3-urllib3
```
vitabaks, line 43: we don't need to install this package by default using the default package manager if the pip installation method is selected urllib3 will be installed along with all other [requirements](https://github.com/vitabaks/postgresql_cluster/blob/master/files/requirements.txt)

I think this package does not have to be explicitly specified even if the rpm/deb installation method is selected since the package manager itself must identify and install the dependent packages.

Or, move this package to the `patroni_packages` section.

vars/Debian.yml
```diff
@@ -94,6 +95,8 @@ installation_method: "repo"  # (default)"repo" or "file"
 # Patroni package will be installed from the pip (by default).
 # You also have the option of choosing an patroni installation method using the deb package.
 patroni_installation_method: "pip"  # (default)"pip" or "deb"
+patroni_packages:
```
vitabaks, line 98:
```suggestion
# if patroni_installation_type: "deb"
patroni_packages:
```

vars/RedHat.yml
```diff
@@ -112,6 +113,9 @@ installation_method: "repo"  # (default)"repo" or "file"
 # Patroni package will be installed from the pip repository (by default).
 # You also have the option of choosing an patroni installation method using the rpm package.
 patroni_installation_method: "pip"  # (default)"pip" or "rpm"
+patroni_packages:
```
vitabaks, line 116:
```suggestion
# if patroni_installation_type: "rpm"
patroni_packages:
```

PR #277, opened 2023-03-09
New role - postgresql-props

mrsrvman:
Consolidation of roles (postgresql-databases,postgresql-users,postgresql-schemas) into one.

Ability to configure user rights and connection limits


vitabaks:
@mrsrvman Hi! Thank you for your contribution to the development.

Please, return the roles to their original state (without consolidation), since these are different tasks. 

Next, you can add the necessary functionality to these roles.

mrsrvman:
Without such consolidation, full user customization may not be possible, which may require the presence of a created base.
It is because of this that you have to create a user, create databases, configure the user (user privileges).

vitabaks:
> Without such consolidation, full user customization may not be possible, which may require the presence of a created base.

there should be no problems if the necessary variables are specified in vars/main.yml

mrsrvman:
Multiple roles do not give the correct execution order.
Merging roles gives you such control.
Variables in the request are unaffected and backwards compatible.

vitabaks:
maybe we need a role "postgresql-privs"

https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_privs_module.html#ansible-collections-community-postgresql-postgresql-privs-module

mrsrvman:
Such a role or task in the role is needed.
There is already a warning when using ansible.postgresql_user.priv.

You may also need to refactor the variables that describe the user and the relationship with the databases.

Now two dictionaries are used - user and bases.
Maybe it makes sense to make one - the user, his bases and possible privileges.

vitabaks:
I think separate variables are needed for privileges

PR #280, opened 2023-03-13
Ability to configure connection limits

mrsrvman:
Replace -  https://github.com/vitabaks/postgresql_cluster/pull/277

### DIFF
--- a/roles/postgresql-databases/tasks/main.yml
+++ b/roles/postgresql-databases/tasks/main.yml
@@ -2,18 +2,19 @@
 
 - name: Make sure the PostgreSQL databases are present
   become: true
-  become_user: postgres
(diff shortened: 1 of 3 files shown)


roles/postgresql-databases/tasks/main.yml
```diff
@@ -2,18 +2,19 @@
 
 - name: Make sure the PostgreSQL databases are present
   become: true
-  become_user: postgres
+  become_user: "{{ patroni_superuser_username }}"
```
vitabaks, line 5: patroni_superuser_username defines the database user, while in become_user we must specify the system user.

patroni_superuser_username may differ from the system user name and there may not be such a system user at all and we will get an error.

vitabaks:
Thanks @mrsrvman

PR #297, opened 2023-04-05
fix(balancers): install requirement packages

wolacinio:
Balancers must install a system_packages:
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'semanage'
fatal: [10.10.10.13]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (libsemanage-python) on ...'s Python /usr/bin/python3. Please read module documentation and install in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
...ignoring

### DIFF
--- a/balancers.yml
+++ b/balancers.yml
@@ -72,6 +72,35 @@
       when: firewall_enabled_at_boot|bool
       tags: firewall
 
+    # Install packages from repository
+    # RedHat
+    - name: Install system packages
+      package:
+        name: "{{ item }}"
+        state: present
+      loop: "{{ system_packages }}"
+      register: package_status
+      until: package_status is success
+      delay: 5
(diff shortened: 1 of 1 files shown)


vitabaks:
It's strange but I don't get such an error when testing.

Please see tasks https://github.com/vitabaks/postgresql_cluster/blob/master/roles/haproxy/tasks/main.yml#L465

instead of installing all the system packages (which are mainly designed for postgres servers and not balancers) we could specify the necessary dependencies in these tasks.

wolacinio:
Ok. Thanks for reply. I check this in next time.

PR #300, opened 2023-04-09
remove_cluster: stop consul service if remove_consul is true

bazzzsh:
Check if remove_consul is true

### DIFF
--- a/remove_cluster.yml
+++ b/remove_cluster.yml
@@ -37,7 +37,7 @@
         name: consul
         state: stopped
         enabled: false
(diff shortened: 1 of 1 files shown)


vitabaks:
thanks @bazzzsh

remove_cluster.yml
```diff
@@ -37,6 +37,7 @@
         name: consul
         state: stopped
         enabled: false
+      when: remove_consul|bool
```
vitabaks, line 40:
```suggestion
      when: remove_consul|bool
```

PR #305, opened 2023-04-22
Add a CONTRIBUTING.md file and Makefile to guide new contributors

### LINKED ISSUE #304: Add a CONTRIBUTING.md file to guide new contributors
Hello @vitabaks,

Firstly, I want to express my appreciation for your amazing work on this project. As the project grows and attracts more contributors, I think it would be helpful to add a CONTRIBUTING.md file to the repository. This file would guide new contributors on how to effectively participate in the project.

Having a CONTRIBUTING.md file can streamline the process for potential contributors and ensure that they have a clear understanding of the steps to follow when submitting issues, making pull requests, and more.

I have already drafted a CONTRIBUTING.md file and will create a pull request for your review. Please let me know if you have any suggestions or changes before I proceed.

Thank you for your time and consideration.

Best regards,

ThomasSanson:
close #304


files/requirements.dev.txt
```diff
@@ -0,0 +1,5 @@
+ansible-lint==6.15.0
```
vitabaks, line 1: why do we fix the version instead of specifying the minimum allowed version? 

The fact is that everything is developing rapidly and very soon these versions will be irrelevant and we will have to periodically maintain this file so that it is up to date.

files/requirements.txt
```diff
@@ -13,3 +13,4 @@ psutil>=2.0.0
 ydiff>=1.2.0
 pexpect>=4.8.0
 ruamel.yaml>=0.16.10
+ansible>=2.7
```
vitabaks, line 16: note that the file requirements.txt it is used to [install](https://github.com/vitabaks/postgresql_cluster/blob/master/roles/patroni/tasks/main.yml#L32) dependencies on the target database servers, I do not think that ansible needs to be installed there, it is not needed there.

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+You can really make a difference by:
+
+- [Making an issue](https://help.github.com/articles/creating-an-issue/). A well-described issue helps a lot. (Check the [known issues](https://github.com/search?q=user%3Avitabaks+is%3Aissue+state%3Aopen) first.)
+- [Making a pull request](https://docs.github.com/fr/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) when you find an error in the code.
+
+I will try to help and take every contribution seriously.
+
+This is a great opportunity for me to learn how you use the role and also an opportunity for you to get into the habit of contributing to open-source software.
+
+## [Step by step](#step-by-step)
```
vitabaks, line 12: There is another way, it is to use gitpod for development, which uses my [image](https://github.com/vitabaks/postgresql_cluster/blob/master/.gitpod.yml#L2) with already installed dependencies.

http://gitpod.io/#https://github.com/vitabaks/postgresql_cluster/pull/305 

Try it and maybe it's worth describing this way ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+Now you can start editing on your laptop.
+
+### [4. Test your changes](#4-optionally-test-your-changes)
+
+Install [make](https://www.gnu.org/software/make/) and [Python3.10](https://www.python.org/) with [venv](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/) :
+
+```
+sudo apt install -y make python3.10 python3.10-venv
```
vitabaks, line 43:
```
gitpod@thomassanso-postgresqlc-9nwq8iqng2j:/workspace/postgresql_cluster$ sudo apt install -y make python3.10 python3.10-venv
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Package python3.10-venv is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'python3.10-venv' has no installation candidate
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+Run `make reinitialization` or `make bootstrap` for initializing the virtualenv and installing all dependencies.
+
+Then, run the test with `make molecule-test` or `make molecule-converge` for only the converge sequence.
+
+If you want to test a specific distribution, set `distro`, optionally `tag`, and `namespace` :
+
+```
+IMAGE_DISTRO=rockylinux9 IMAGE_TAG=latest IMAGE_NAMESPACE=geerlingguy make molecule-converge
```
vitabaks, line 63:
```
gitpod@thomassanso-postgresqlc-9nwq8iqng2j:/workspace/postgresql_cluster$ IMAGE_DISTRO=rockylinux9 IMAGE_TAG=latest IMAGE_NAMESPACE=geerlingguy make molecule-converge
/bin/bash: line 1: .venv/bin/activate: No such file or directory
Traceback (most recent call last):
  File "/usr/local/bin/molecule", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/command/converge.py", line 59, in converge
    base.execute_cmdline_scenarios(scenario_name, args, command_args, ansible_args)
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/command/base.py", line 94, in execute_cmdline_scenarios
    get_configs(args, command_args, ansible_args, glob_str),
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/command/base.py", line 185, in get_configs
    configs = [
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/command/base.py", line 186, in <listcomp>
    config.Config(
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/config.py", line 64, in __call__
    obj.after_init()
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/config.py", line 118, in after_init
    self._validate()
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/config.py", line 456, in _validate
    errors = schema_v3.validate(self.config)
  File "/home/gitpod/.local/lib/python3.10/site-packages/molecule/model/schema_v3.py", line 61, in validate
    with open(schema_file, encoding="utf-8") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.10/dist-packages/molecule_docker/driver.json'
make: *** [.config/make/molecule.mak:14: molecule-converge] Error 1
```

files/requirements.dev.txt
ThomasSanson: You're right. To help manage dependencies and keep them up-to-date, developers can use automated tools like Renovate or Dependabot. These tools monitor the dependencies specified in your project and automatically create pull requests to update the versions when new releases become available. By integrating these tools into your development workflow, you can achieve the following benefits:

1. Up-to-date dependencies: Renovate and Dependabot help you stay current with the latest versions of your dependencies, ensuring you receive the benefits of new features, performance improvements, and security patches.

2. Reduced maintenance burden: Automating the update process can save developers time and effort spent on maintaining the dependencies manually. The tools handle the updates, allowing the developers to focus on other aspects of the project.

3. Compatibility checks: When a new dependency version is available, these tools create a pull request with the updated version, triggering your CI/CD pipeline. This process helps catch potential compatibility issues or conflicts with the new version before merging the changes into your main codebase.

4. Customizable configuration: You can configure the update frequency, versioning strategy (e.g., semantic versioning or pinned versions), and other settings to suit your project's requirements.

By combining the use of Renovate or Dependabot with specifying minimum allowed versions or version ranges in your requirements file, you can enjoy the advantages of both up-to-date dependencies and automated maintenance while reducing the risk of introducing breaking changes or compatibility issues.

files/requirements.txt
ThomasSanson: I apologize for my mistake. I misunderstood the purpose of the requirements.txt file and thought it was for the Ansible bastion. It seems that it's used to install dependencies on the target database servers, and you're right that Ansible doesn't need to be installed there.

I will make the necessary corrections tomorrow to resolve this issue.

CONTRIBUTING.md
ThomasSanson: It seems that the command `sudo apt update` is missing in the current configuration, which is necessary for proper functionality. I will add this command to ensure everything works as intended.

CONTRIBUTING.md
ThomasSanson: The error encountered is expected, as the package python3.10-venv couldn't be installed, which prevents the creation of the Python virtual environment. This issue will be resolved once the missing command is added earlier in the process, allowing for the proper installation of python3.10-venv and successful creation of the virtual environment.

files/requirements.txt
ThomasSanson: This situation highlights another reason to consider switching to collections😄. But on a more serious note, I understand the need to keep the configuration as simple and "KISS" (Keep It Simple, Stupid) as possible, and I will take this into account moving forward.

CONTRIBUTING.md
ThomasSanson: Indeed, using Gitpod seems to be a better option overall!

In any case, having a CONTRIBUTING.md file is important for guiding contributors.

Initially incorporating a Makefile (and possibly Taskfile later) is a good idea, as many open-source projects are built this way, and it allows people to understand the workflow more easily.

Another reason, which I find significant but is open to discussion, is that we're not restricted to a single dependency like Gitpod. The Makefile and virtual environment can work anywhere (VDI, Docker, Gitpod, etc.).

Please note, I'm a big fan of using Gitpod as well!

I will add the Gitpod section to the CONTRIBUTING.md file because it's a great addition.

I'll also test with your Docker image to see if I can incorporate the virtual environment directly. This way, through Gitpod, we can easily run a `make molecule-test` if needed.

What are your thoughts on this?

CONTRIBUTING.md
ThomasSanson: Done and tested on gitpod (working)

CONTRIBUTING.md
ThomasSanson: Done and tested on gitpod (working)

files/requirements.txt
ThomasSanson: Done and tested on gitpod with Makefile (working)

CONTRIBUTING.md
ThomasSanson: I think I need to update the Dockerfile to install Python 3.10 and necessary tools by default. However, I would prefer to use Gitpod with the "tasks" method and initialize it via the makefile.

In summary:
- Remove the "# install molecule and ansible" section from Dockerfile.
- In .gitpod.yml, under tasks:
  - Under the "init" task, include the following commands:
     ```
     sudo apt update
     sudo apt install -y make python3.10 python3.10-venv
     python3 -m pip install --upgrade pip
     python3 -m pip install virtualenv
     ```
  - Under the "command" task, include `make reinitialization`.

This way, Gitpod will be consistent with the local setup (without Gitpod).

What do you think?

files/requirements.dev.txt
vitabaks: or just use the latest versions in tests. Or do you think that this approach is wrong?

files/requirements.txt
vitabaks: the downside of this approach is the increase in the creation time of the gitpod working environment. 

Do we really need compliance with gitpod? Why can't we consider this as two independent options?

CONTRIBUTING.md
vitabaks: the downside of this approach is the increase in the creation time of the gitpod working environment. 

Do we really need compliance with gitpod? 
Why can't we consider this as two independent options?

CONTRIBUTING.md
```diff
@@ -0,0 +1,127 @@
...
+6. **Commit your changes**: Once you've made your changes, stage and commit them using the following commands:
+
+   ```
+   git add .
+   git commit -m "Add a brief description of your changes"
+   ```
+
+   Replace the commit message with a clear and concise description of your changes.
+
+7. **Push your changes**: Push your changes to your fork on GitHub:
```
vitabaks, line 70: It seems that the "Test your changes" item is missing here before pushing changes

CONTRIBUTING.md
ThomasSanson: I understand your concern about the increased creation time of the Gitpod working environment. However, the goal here is to provide a consistent experience for developers, whether they choose to work with Gitpod or on their local setup. This consistency can help reduce confusion and streamline the development process.

If we want to minimize the creation time, we should consider making similar optimizations to the Dockerfile, so that the same efficiency is achieved in both Gitpod and local setups. This way, developers will have the flexibility to choose their preferred environment without any major trade-offs in terms of setup time.

For example, we can update the Dockerfile to include Python 3.10 and the necessary tools by default, similar to what we did in the Gitpod configuration.

By doing this, we can maintain consistency between Gitpod and local environments, while also minimizing the creation time.

Let me know your thoughts on this.

ThomasSanson: I would like to emphasize an important point for me. I believe it is crucial to maintain consistent functionality between Gitpod and local installations. Moreover, it is essential that people can quickly join the project and get started with ease.

Gitpod is an excellent choice, but I think it is important not to become too reliant on it. And what about those who need to be "offline" or on a private network?

That's why I think the Makefile approach, which would serve as a link, seems like a good choice.

Now, maybe you prefer that people learn to use Molecule and Python virtual environments? Perhaps it would be better to provide a basic Gitpod environment, such as the Ubuntu one you have, and add step-by-step installation instructions (the requirements.yml file remains essential to me). This way, new contributors can install in the same manner and have the same testing and contribution instructions, whether they start from the Gitpod base or their local machine.

What are your thoughts on this?

files/requirements.dev.txt
ThomasSanson: Based on my experience with linters, if we do not freeze their versions, they can cause errors either during the automatic CI pipelines or while running tests. This can be quite disruptive for many of my colleagues when they start working on a specific task (let's say adding a new feature) and then find themselves dealing with updates from the linters.

That's why we choose to fix the versions of all the tools we use (such as Ansible, ansible-lint, yamllint, etc.) and create specific issues to address their updates.

As updating these tools can be tedious, we've automated the process using Renovate, which gradually updates the versions while running all the tests.

This way, we can focus on addressing only the non-trivial (blocking) cases, making our workflow more efficient and less frustrating for the team.

CONTRIBUTING.md
vitabaks: good. If you have time, please provide a PR for gitpod based on how you see it.

But I think it is important that gitpod is already fully configured when starting the production environment.

It is necessary to save time, because it is the most valuable resource)

CONTRIBUTING.md
vitabaks: good. If you have time, please provide a PR for gitpod based on how you see it.

But I think it is important that gitpod is already fully configured when starting the production environment.

It is necessary to save time, because it is the most valuable resource)

files/requirements.dev.txt
ThomasSanson: I've addressed the issue with the molecule version in the Makefile. That's why it's best to pin versions and update them gradually as needed.

Additionally, having a more unified approach for testing is beneficial.

I'm adding all of this here.

I know I'm advocating for my own perspective, but in any case, you remain the authority in validating this PR, so please don't hesitate to provide feedback. I don't want to come across as overly intrusive!

CONTRIBUTING.md
ThomasSanson: Done and tested on gitpod (working)

ThomasSanson:
@vitabaks 

I've made all the necessary changes for the contribution to ensure that there is maximum consistency between a local installation and Gitpod via the Makefile.

Indeed, there is a difference in launch time. For you, the Gitpod workspace is the most efficient at 30.77 seconds. With the current configuration file, it takes 56.8 seconds for the first image build and then 53.41 seconds for the virtual environment initialization (make bootstrap), which results in a total of 110.21 seconds for the first startup, almost four times longer.

However, if I delete my Gitpod workspace and recreate it, the image is already built, and I only have the approximately 53 seconds for the opening and make bootstrap process.

And if I turn off my workspace and turn it back on, it takes 32.29 seconds (very close to your result).

I'm torn between a maximum time of 2 minutes for a brand new contributor, who will then get closer to your times but with consistency between installations, and pure performance.

I'll leave the decision to you 😇

files/requirements.dev.txt
vitabaks: OK I agree with you

vitabaks:
> Is there a reason for using your pre-built Docker image directly? (Faster?)

In this case, there will be a new build for each workspace launch. Yes, you can wait a bit for the build, but if you use an already pre-assembled image, the workspace will be launched much faster.

PR #307, opened 2023-04-27
Integrate code improvements and add linters

### LINKED ISSUE #306: Integrate code improvements and add linters
I have identified several potential improvements in the code and would like to integrate them. These improvements are described below:

1. Add bash-completion package and sort packages
- Add the bash-completion package to the .gitpod.Dockerfile file to enable tab completion in the terminal.
- Sort the packages in the .gitpod.Dockerfile file.

2. Clean up and fix indentation in the molecule.yml file
- Fix indentation issues to improve the readability of the molecule.yml file.

3. Add linters for YAML, Ansible, and Python files
- Update the .yamllint file to ignore the .venv directory and enforce a minimum of one space between comments and content.
- Update the .ansible-lint file to exclude the .venv directory.
- Update the Makefile to include the new linters.mak file, which contains the targets for the linters.
- Run the linters in the `lint` target.

To resolve this issue, I will submit a Pull Request with the above changes.

ThomasSanson:
close #306


molecule/postgrespro/converge.yml
```diff
@@ -26,8 +26,6 @@
         - ansible_distribution_major_version is version('8', '>=')
 
     - name: Include postgrespro vars
-      include_vars: ./postgrespro_vars.yml
+      ansible.builtin.include_vars: vars/postgrespro_vars.yml
```
vitabaks, line 29:
```suggestion
      include_vars: vars/postgrespro_vars.yml
```

The `ansible.builtin` prefix was introduced in Ansible 2.10. In this version, Ansible's collections and modules structure underwent significant changes. Many modules were moved out of the main Ansible codebase into separate collections, and the ansible.builtin. prefix was introduced for modules that are still part of the main Ansible distribution.

For Ansible versions prior to 2.10, you can simply use the module name without the ansible.builtin. prefix. If you want to ensure backward compatibility with older Ansible versions, you can continue to use the short module names without the prefix.

CONTRIBUTING.md
```diff
@@ -37,9 +37,10 @@ Use Gitpod for a cloud-based development environment:
 4. Create a new branch: `git checkout -b my-feature-branch`
 5. Make your changes and commit: `git add .` and `git commit -m "Description of changes"`
 6. Test with Molecule: `make molecule-test` or `make molecule-converge`
-7. Push your changes: `git push origin my-feature-branch`
-8. Create a pull request on GitHub
-9. Wait for a review
+7. Test with linters(otpional): `make lint`
```
vitabaks, line 40:
```suggestion
7. Test with linters: `make lint`
```

vitabaks: I would recommend to run the linters first and only then the tests.

molecule/postgrespro/converge.yml
ThomasSanson: Actually, I initially tried to fix a linter issue by adding `ansible.builtin`, but it turned out that the problem was related to the file location rather than `ansible.builtin`. Then I moved the `postgrespro_vars.yml` file but forgot to remove `ansible.builtin`. So, I will fix that.

Furthermore, since we use virtualenv and plan to switch to collection in the future, we will need to use a minimum version of Ansible 2.11. We will discuss this again at the time of the transition to collection, but that is not expected to happen anytime soon.

CONTRIBUTING.md
ThomasSanson: I completely agree that linters should be integrated into tests and made mandatory, meaning that code cannot be validated and pushed unless it passes the linter check. However, since we are not yet fully clean on the linters, I think it's better to leave them optional for now and personally check the pull requests.

My belief is that it's the responsibility of the projects (and therefore the project managers) to ensure that the necessary tools are in place for contributors. Therefore, I suggest that we leave the linters optional for the time being and make them mandatory directly in the Makefile once the work on the linters is done.

What do you think of this approach?

PR #365, opened 2023-06-06
Add a GitHub Action for Docker (make docker-build)

### LINKED ISSUE #353: Add a GitHub Action for Docker (make docker-build)
Hello,

As we know, Docker is a key part of our infrastructure, and we need to ensure that changes to the repository do not inadvertently break the Docker build process. To this end, I propose that we add a GitHub Action that will automatically execute `make docker-build` whenever changes are pushed to the repository. This will help us catch potential issues early and ensure that our Docker environment remains stable and reliable.

This action can be set up in a new workflow file under the `.github/workflows` directory, and it would look something like this:

```yaml
name: Docker Build Test

on: [push, pull_request]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2

    - name: Docker Build Test
      run: make reinitialization && make docker-build
```

This setup will run the `make docker-build` command on every push or pull request, ensuring that our Docker build is always functional.

Please let me know your thoughts on this. If there's consensus on the idea, I'd be happy to take the lead on implementing this feature.

Best,

ThomasSanson:
Close #353

### DIFF
--- a/.github/workflows/gitpod.yml
+++ b/.github/workflows/gitpod.yml
@@ -1,0 +1,32 @@
+---
+name: "Gitpod"
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Set TERM environment variable
+        run: echo "TERM=xterm" >> $GITHUB_ENV
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Install dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/gitpod.yml
```diff
@@ -0,0 +1,32 @@
+---
+name: "Gitpod"
```
vitabaks, line 2:
```suggestion
name: "Docker"
```

I'm not sure we should specify the name "Gitpod" here.

ThomasSanson: Thank you for your feedback. I was indeed uncertain about this. Although we are using Docker, it is currently confined to Gitpod. However, you're right, we should avoid premature optimization. For now, I'm going to change the name to "Docker". We'll reconsider this decision when we have more Docker implementations. Thanks again for your constructive input.

PR #367, opened 2023-06-07
feat: add tests for Ansible Verifier

ThomasSanson:
🚀 feat(molecule): add Postgres and Patroni tests
✨ feat(molecule): add etcd health check test
🐛 fix(molecule): enable verifier in molecule config The molecule configuration file was updated to enable the verifier. A new group_vars file was added to set the default Postgres database name, user, and password. Tests were added to check if PostgreSQL is running, listening on the default port, and if the database exists. A test was added to check the replication status of PostgreSQL. Tests were added to check the status of Patroni and etcd. These tests will help ensure that the application is running correctly and that the infrastructure is properly configured.

### DIFF
--- a/molecule/default/group_vars/all.yml
+++ b/molecule/default/group_vars/all.yml
@@ -1,4 +1,0 @@
----
-postgres_db_name: postgres
-postgres_user: postgres
-postgres_password: postgres
--- a/molecule/default/group_vars/all.yml
+++ b/molecule/default/group_vars/all.yml
@@ -1,0 +1,4 @@
+---
+postgres_db_name: postgres
+postgres_user: postgres
+postgres_password: postgres
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -6,9 +6,9 @@
     - name: Molecule | Verify | Include all tests
       ansible.builtin.include_tasks: "{{ item }}"
       with_fileglob:
-        - ../tests/postgres/postgres.yml
-        - ../tests/patroni/*.yml
         - ../tests/etcd/*.yml
+        - ../tests/patroni/*.yml
+        - ../tests/postgres/postgres.yml
 
 - name: Molecule | Verify | Replica
   hosts: replica
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -5,7 +5,7 @@
   tasks:
     - name: Include vars of vars/main.yml
       ansible.builtin.include_vars:
-        file:  ../../vars/main.yml
+        file: ../../vars/main.yml
 
     - name: Molecule | Verify | Include all tests
       ansible.builtin.include_tasks: "{{ item }}"
--- a/molecule/default/verify.yml
+++ b/molecule/default/verify.yml
@@ -3,18 +3,13 @@
   hosts: all
 
   tasks:
+    - name: Include vars of vars/main.yml
+      ansible.builtin.include_vars:
+        file:  ../../vars/main.yml
+
     - name: Molecule | Verify | Include all tests
       ansible.builtin.include_tasks: "{{ item }}"
       with_fileglob:
         - ../tests/etcd/*.yml
         - ../tests/patroni/*.yml
-        - ../tests/postgres/postgres.yml
-
-- name: Molecule | Verify | Replica
-  hosts: replica
-
-  tasks:
-    - name: Molecule | Verify | Include replica tests
-      ansible.builtin.include_tasks: "{{ item }}"
-      with_fileglob:
-        - ../tests/postgres/replication.yml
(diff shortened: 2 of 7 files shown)


vitabaks:
Thank you for your contribution 

I think we should clarify the title of this PR
In my understanding, there are already tests - https://github.com/vitabaks/postgresql_cluster/actions/workflows/molecule.yml

vitabaks:
I believe we are talking about additional tests here

It is important to note that the necessary tests are performed in the main playbook

here is an example:

https://github.com/vitabaks/postgresql_cluster/blob/master/roles/patroni/tasks/main.yml#L836
https://github.com/vitabaks/postgresql_cluster/blob/master/roles/etcd/tasks/main.yml#L122

You can find a lot of detailed checks in the code.

vitabaks:
maybe so "Add tests for Ansible Verifier"

molecule/default/verify.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: Molecule | Verify | All
+  hosts: all
+
+  tasks:
+    - name: Molecule | Verify | Include all tests
+      ansible.builtin.include_tasks: "{{ item }}"
+      with_fileglob:
+        - ../tests/postgres/postgres.yml
+        - ../tests/patroni/*.yml
+        - ../tests/etcd/*.yml
```
vitabaks, line 11:
```suggestion
        - ../tests/etcd/*.yml
        - ../tests/patroni/*.yml
        - ../tests/postgres/postgres.yml
```

it is worth checking etcd first and then patroni and only then postgres

molecule/tests/postgres/postgres.yml
```diff
@@ -0,0 +1,23 @@
...
+  command: pgrep -u postgres
+  register: result
+  failed_when: result.rc != 0
+
+- name: Check if PostgreSQL is listening on the default port
+  wait_for:
+    port: 5432
+    timeout: 5
+  register: is_listening
+  failed_when: not is_listening
```
vitabaks, line 12:
```suggestion
- name: Check if PostgreSQL is listening on the port {{ postgresql_port }}
  wait_for:
    port: "{{ postgresql_port }}"
    timeout: 5
  register: is_listening
  failed_when: not is_listening
```

molecule/tests/postgres/postgres.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Try to connect to PostgreSQL
+  postgresql_ping:
+    login_user: "{{ postgres_user }}"
+    login_password: "{{ postgres_password }}"
+    db: postgres
```
vitabaks, line 18:
```suggestion
    db: "{{ postgres_db_name }}"
```

molecule/tests/postgres/postgres.yml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Check if the database exists
```
vitabaks, line 20: if we have previously connected to this database, then it is there. 

It may be necessary to add an additional database to the deployment tests by defining it in the postgresql_databases variable

molecule/tests/postgres/postgres.yml
ThomasSanson: Indeed, that aligns with my initial intention. Firstly, I would like to test access to the Postgres database. Would it be acceptable at this stage to directly connect to the Postgres database? It should always be there, right? Afterwards, we will test access to another database.

However, as you've pointed out, this requires creating additional databases through `postgresql_databases`. I'll make these corrections.

molecule/tests/postgres/postgres.yml
ThomasSanson: see https://github.com/vitabaks/postgresql_cluster/pull/367#discussion_r1222818273

vitabaks:
It doesn't sound bad and it will take a lot of effort. Thanks!

vitabaks:
Thank you for sharing your detailed testing approach. 
I find the idea of integrating both installation testing and post-installation verification compelling. The planned 'healthcheck' playbook and future 'side_effect' implementation certainly add robustness to the process. 
Looking forward to seeing these enhancements in action.

vitabaks:
@ThomasSanson Let me know when you're done working on this PR.

molecule/default/group_vars/all.yml
```diff
@@ -0,0 +1,4 @@
+---
+postgres_db_name: postgres
+postgres_user: postgres
+postgres_password: postgres
```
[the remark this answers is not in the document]
ThomasSanson: Done

molecule/tests/postgres/postgres.yml
ThomasSanson: Done

PR #370, opened 2023-06-09
Fix linter for Consul

ThomasSanson:


ThomasSanson:
Hello @vitabaks,

I have a question regarding the Consul role. I was wondering if this role has been tested with Molecule ?

Based on my own investigation, I didn't find indications that this is the case, but I wanted to make sure I'm not missing something. Could you confirm whether this is correct or not ?

Thank you in advance for your time and your help.

vitabaks:
@ThomasSanson That's right, I haven't added this scheme to the Molecule tests yet.

I have an idea to add a consul schema test to the daily tests (schedule_pg_xxx.yml). 

Where one version of the Linux distribution will be deployed in the scheme with Etcd the second version of Consul.

What do you think?

ThomasSanson:
@vitabaks,

I agree with your suggestion to add a consul schema test to the daily tests. However, I believe we could also benefit from adding Molecule scenarios, just like what has been done for the default and postgrespro scenarios.

Thus, I would lean towards adding a "consul" or Type C scenario. These scenarios would be very helpful in testing and verifying different configurations and environments.

Regarding the Github Action Matrix, I will need to look into it. Without it, maintaining multiple scenarios could become challenging.

We can merge this pull request if you wish. Afterward, we could open a new issue to discuss and work on adding the Type C scenario. This will keep the discussions organized and make progress tracking easier.

PR #371, opened 2023-06-09
Fix yaml[truthy] from all workflow files

ThomasSanson:
The yaml[truthy] rule was removed from the ansible-lint configuration file, so it is no longer necessary to exclude it from the workflows. This commit removes the yaml[truthy] exclusion from all workflow files.

### DIFF
--- a/.config/ansible-lint.yml
+++ b/.config/ansible-lint.yml
@@ -25,10 +25,10 @@ skip_list:
   - yaml[line-length]
   - yaml[octal-values]
   - yaml[trailing-spaces]
-  - yaml[truthy]
 
 exclude_paths:
   - ../roles/consul/ # TODO - https://github.com/ansible-community/ansible-consul/pull/520
   - ../.venv
+  - ../.github
 # https://ansible-lint.readthedocs.io/configuring/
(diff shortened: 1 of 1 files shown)


ThomasSanson:
Hello @vitabaks,

I've been giving some thought to the files in the .github directory. Since these files are not part of Ansible, I'm wondering if it might be better to ignore them when running ansible-lint, and just use yamllint instead.

What are your thoughts on this? Do you think it would be a good idea, or is there a reason for including these files in ansible-lint that I may have missed?

Looking forward to your feedback.

vitabaks:
That's right

ansible-lint needs to be performed only for playboys and ansible roles.

.github/workflows/docker.yml
```diff
@@ -1,7 +1,7 @@
 ---
 name: "Docker"
 
-on:
+on: # noqa yaml[truthy]
```
vitabaks, line 4: I suggest removing this comments and simply excluding the `.github` directory from the ansible-lint checks.

ThomasSanson: Done

PR #372, opened 2023-06-09
Fix linter yaml octal values

ThomasSanson:


ThomasSanson:
Darn, it appears the modifications from yaml[truthy] have also been included

Issue #374, opened 2023-06-09
Implement retry logic in Docker workflow to handle network failures


ThomasSanson:
Hello,

We have been experiencing occasional failures in our Docker GitHub Actions workflow due to network issues. These failures usually occur in the "Run Docker tests" steps of the workflow.

To mitigate this, I would like to implement a retry mechanism in our workflow. The idea is to retry these steps a few times (let's say 3 times) before marking the workflow as failed. This should help us handle transient network issues.

Considering using the `nick-invision/retry@v2` action for this, which would modify our workflow file to look like this:

```yaml
name: "Docker"

on:  
  push:  
    branches:  
    - master  
  pull_request:  
    branches:  
    - master  

jobs:  
  build:  
    runs-on: ubuntu-latest  
    steps:  
    - name: Set TERM environment variable  
      run: echo "TERM=xterm" >> $GITHUB_ENV  

    - name: Checkout  
      uses: actions/checkout@v3  

    - name: Set up Python 3.10  
      uses: actions/setup-python@v4  
      with:  
        python-version: "3.10"  

    - name: Install dependencies  
       run: make bootstrap-dev  

    - name: Run Docker tests  
      uses: nick-invision/retry@v2
      with:
        max_attempts: 3
        timeout_minutes: 10
        command: make docker-tests
```

https://github.com/vitabaks/postgresql_cluster/actions/runs/5224946567/jobs/9433782895?pr=373#step:6:999

PR #378, opened 2023-06-10
Removal of fqcn[action] rule from ansible-lint skip_list

ThomasSanson:
This pull request introduces several updates and improvements to the project:

1. The `fqcn[action]` rule has been removed from the `skip_list` as it is no longer necessary.

2. The Ansible modules used in the playbook have been updated to use the `community.general` and `community.docker` collections. These modules have been moved to these collections and are no longer available in the core Ansible distribution. This change ensures that the playbook remains compatible with the latest version of Ansible.

3. Deprecated Ansible modules have been replaced with their updated counterparts. This change improves the maintainability of the roles, as the updated modules are more reliable and have better support.

4. The Ansible module names used in the roles have been updated to use fully qualified names. This helps to avoid any potential conflicts with other modules


.config/ansible-lint.yml
```diff
@@ -3,7 +3,6 @@ skip_list:
   - command-instead-of-module # Using command rather than module where necessary
   - command-instead-of-shell # Using command rather than shell where necessary
   - experimental # all rules tagged as experimental
-  - fqcn[action]
```
vitabaks, line 6: I have been trying for a long time to ensure compatibility with older versions of ansible, but it seems it's time to raise the minimum version to 2.11, for example, because now there is no compatibility with older versions.

Please mark the minimum version in `minimal_ansible_version` variable and in the README.

vitabaks:
I think within the framework of this PR, it is now necessary to set the fqdn for all modules throughout the code, which of course is a lot of time.

@ThomasSanson Do you need my help with this?

ThomasSanson:
@vitabaks,

Thank you for your offer, but I'll handle this and gradually address the skip_list of ansible-lint.

Regardless, this pull request will be approved. However, I must first change the minimum version of Ansible.

.config/ansible-lint.yml
ThomasSanson: Done, pull request is ready

ThomasSanson:
@vitabaks,

Could you please restart the GitHub Docker action ?

See #374

ThomasSanson:
pull request is ready 😇

vitabaks:
I skimmed through the code and found that there are still a lot of modules without fqdn. Please tell me why you translated only part of the code to fqdn and not the whole code? 

I understand that if we decided to use this style, then it should be followed throughout the code. 

For example
instead of `apt` should be  `ansible.builtin.apt`
instead of  `debug` should be `ansible.builtin.debug`

Or do I understand something wrong?

ThomasSanson:
I activated only `fqcn[action]` and not `fqcn[action-core]` because it allows me to make progress gradually. 

I will handle the other part in a separate pull request. 

This way, it reduces the workload for your reviews too.

vitabaks:
thank you for the clarification.

I completely agree with you, it is much more convenient to go gradually in small steps.

PR #387, opened 2023-06-29
Dockerfile Optimization with Multi-Stage Builds

### LINKED ISSUE #359: Proposal for Dockerfile Optimization with Multi-Stage Builds
Hello,

I have been examining the Dockerfile located in `.config/gitpod/Dockerfile` of the `postgresql_cluster` project and I would like to propose an enhancement using multi-stage builds. 

Multi-stage builds are a feature in Docker that allow you to structure your Dockerfile into multiple stages. Each stage is like a mini-Dockerfile with its own `FROM` statement. The final stage can selectively copy artifacts from the previous stages, leaving behind everything not explicitly copied. This allows you to separate the build-time dependencies from the runtime dependencies, reducing the final image size and attack surface, making your containers more secure and efficient.

The primary benefits of multi-stage builds include:

1. **Reduced Image Size**: Only the artifacts that are needed at runtime are copied to the final image. This reduces the size of the image by leaving out the build tools and intermediate files that are not needed in production.

2. **Improved Readability and Maintainability**: By splitting the Dockerfile into multiple stages, it becomes easier to read and maintain. Each stage has a clear purpose, which simplifies understanding and debugging.

3. **Faster Build Times**: Docker can cache the results of each stage separately. This means that if you change something in one stage, Docker only needs to rebuild that stage and the ones after it.

4. **Reduced Attack Surface**: By minimizing the contents of the final image, you also minimize the potential attack surface for a malicious actor. This makes your container more secure.

Here's how the Dockerfile might look with multi-stage builds:

```Dockerfile
# First stage: Install build-time dependencies
FROM ubuntu:jammy as builder

USER root

# Copy Python version config file
COPY .config/python_version.config /tmp/

# Set a variable for the npm version
ARG NPM_VERSION=9.6.7
# Set a variable for the ungit version
ARG UNGIT_VERSION=1.5.23

# Update system and install packages*
# hadolint ignore=DL3008,DL3013
RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) \
    && apt-get update \
    && apt-get upgrade -y \
    && apt-get install -y --no-install-recommends \
        bash-completion \
        ca-certificates \
        curl \
        git \
        git-lfs \
        gnupg \
        htop \
        iproute2 \
        lsb-release \
        make \
        nano \
        python3-pip \
        "python${PYTHON_VERSION}" \
        "python${PYTHON_VERSION}-venv" \
        sudo \
        tree \
        vim \
        wget \
    && python3 -m pip install --no-cache-dir --upgrade pip \
    && python3 -m pip install --no-cache-dir virtualenv

# Install Docker
FROM builder as docker-installer

RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
    && echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/docker-archive-keyring.gpg] https://download.docker.com/linux/ubuntu \
    $(lsb_release -cs) stable" | tee /etc/apt/sources.list.d/docker.list > /dev/null \
    && apt-get update \
    && apt-get install -y --no-install-recommends docker-ce docker-ce-cli containerd.io

# Install npm and ungit
FROM docker-installer as npm-installer

RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
    && apt-get install  -y --no-install-recommends nodejs \
    && npm install -g npm@${NPM_VERSION} \
    && npm install -g \
        ungit@${UNGIT_VERSION}

# Cleanup
FROM npm-installer as cleanup

RUN apt-get clean && rm -rf /var/lib/apt/lists/* tmp/* 

# Final stage: Setup final image and user
FROM ubuntu:jammy

COPY --from=cleanup / /

# Create the gitpod user. UID must be 33333.
RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod

USER gitpod
```

By applying this structure, we can make the Dockerfile more efficient and readable, and it reduces the size of the final image by leaving out unnecessary build-time dependencies.

Would you be open to considering this change ? I would be happy to assist further with this enhancement.

Best,

AurelienToitot:
close #359


.config/gitpod/Dockerfile
```diff
@@ -38,23 +36,37 @@ RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) && \
...
     && python3 -m pip install --no-cache-dir --upgrade pip \
-    && python3 -m pip install --no-cache-dir virtualenv \
-    # Install Docker
-    && curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
+    && python3 -m pip install --no-cache-dir virtualenv
+
+# Install Docker
+FROM builder as docker-installer
+
+RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | gpg --dearmor -o /usr/share/keyrings/docker-archive-keyring.gpg \
```
ThomasSanson, line 44: # hadolint ignore=DL3008,DL3009

DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`: This is a best practice to ensure build reproducibility to pin the versions of installed packages. However, this can make your Dockerfile more difficult to maintain. If you want to ignore this warning, you can add `# hadolint ignore=DL3008` above the RUN lines.

DL3009 info: Delete the apt-get lists after installing something: You can delete package lists after installation to reduce image size. You are already doing this in the cleanup phase, but hadolint is likely raising this warning for the apt-get update commands in the intermediate stages. You can ignore this warning.

ThomasSanson, line 44: DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check: For enhanced security, you can add `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` before RUN commands that use pipes.

.config/gitpod/Dockerfile
```diff
@@ -38,23 +36,37 @@ RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) && \
...
     && apt-get update \
-    && apt-get install -y --no-install-recommends docker-ce docker-ce-cli containerd.io \
-    # Install npm and ungit using the versions specified in the variables
-    && curl -fsSL https://deb.nodesource.com/setup_20.x |bash - \
+    && apt-get install -y --no-install-recommends docker-ce docker-ce-cli containerd.io
+
+# Install npm and ungit
+FROM docker-installer as npm-installer
+
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash - \
```
ThomasSanson, line 53: DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check: For enhanced security, you can add `SHELL ["/bin/bash", "-o", "pipefail", "-c"]` before RUN commands that use pipes.

ThomasSanson, line 53: `# hadolint ignore=DL3008`

DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`: This is a best practice to ensure build reproducibility to pin the versions of installed packages. However, this can make your Dockerfile more difficult to maintain. If you want to ignore this warning, you can add `# hadolint ignore=DL3008` above the RUN lines.

.config/gitpod/Dockerfile
```diff
@@ -38,23 +36,37 @@ RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) && \
...
     && apt-get install  -y --no-install-recommends nodejs \
     && npm install -g npm@${NPM_VERSION} \
```
ThomasSanson, line 55: SC2086 info: Double quote to prevent globbing and word splitting.: To prevent glob expansion and word splitting, you can add double quotes around variables in RUN commands. For example, `"npm@${NPM_VERSION}"` instead of npm@${NPM_VERSION}.

.config/gitpod/Dockerfile
```diff
@@ -38,23 +36,37 @@ RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) && \
...
+# Final stage: Setup final image and user
+FROM ubuntu:jammy
+
+COPY --from=cleanup / /
 
 # Create the gitpod user. UID must be 33333.
 RUN useradd -l -u 33333 -G sudo -md /home/gitpod -s /bin/bash -p gitpod gitpod
 
-USER gitpod
+USER gitpod
```
ThomasSanson, line 72: There is a missing newline (or line break) at the end of the file.

This is often mentioned because the POSIX standard defines a line as a sequence of zero or more non-newline characters, followed by a newline character. Therefore, by convention, many tools (including git) expect a file to end with a newline. If this end line is missing, some tools may misinterpret the last line of the file, or display warnings that the file does not end with a newline.

.config/gitpod/Dockerfile
```diff
@@ -38,23 +36,37 @@ RUN PYTHON_VERSION=$(cut -d '=' -f 2 /tmp/python_version.config) && \
...
-    && apt-get clean && rm -rf /var/lib/apt/lists/* tmp/*
+        ungit@${UNGIT_VERSION}
+
+# Cleanup
+FROM npm-installer as cleanup
+
+RUN apt-get clean && rm -rf /var/lib/apt/lists/* tmp/* 
+
+# Final stage: Setup final image and user
+FROM ubuntu:jammy
```
ThomasSanson, line 65: To avoid including the layers from the ubuntu:jammy image in our final image and considering that you're copying the root from the last stage instead of the ubuntu:jammy root, it would be preferable to start from the scratch image.

`FROM scratch`

https://hub.docker.com/_/scratch

.config/gitpod/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM ubuntu:jammy
+# First stage: Install build-time dependencies
```
ThomasSanson, line 1: Congratulations on this contribution! Thank you very much. 

Have you carried out tests using the `make docker-tests` command, in particular ? 

Also, don't hesitate to run `gp validate` to ensure it works correctly on Gitpod

ThomasSanson (changes requested):

.config/gitpod/Dockerfile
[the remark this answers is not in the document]
AurelienToitot: @ThomasSanson

Indeed, following the best coding practice, the idea of running linting before building is much better.

Therefore, I have just made the correction in this commit ef5212363b5cbda73322de5951785bbb53cf0c88.

.config/gitpod/Dockerfile
```diff
@@ -1,9 +1,8 @@
...
+FROM ubuntu:jammy as builder
 
 USER root
```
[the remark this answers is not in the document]
AurelienToitot: @ThomasSanson 

Following your thread, I understand that for security reasons, it is necessary to ignore the hadolint DL3002. Therefore, I made the modification in the Dockerfile, and indeed, the error of USER: root no longer appears.

You will find my correction in this commit: 327841197f4531d325635d5d42314221b742ad72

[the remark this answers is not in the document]
ThomasSanson: @AurelienToitot 
Thank you for the correction. However, please be aware of a comprehension issue. For security reasons, it is recommended not to run the final Docker image with the root user, but with a different user.

In this case, the root user is used in the initial build stage, and in the final stage (which inherits from the previous stage), the user is set as "gitpod". As a result, the final Docker image will not run with the root user, but with the "gitpod" user. So, everything is fine.

The linter hadolint gives a false positive in this case. That's why we can ignore it.

However, it is not true to say that for security reasons, we should ignore it.

That's the confusion.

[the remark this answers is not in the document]
AurelienToitot: @ThomasSanson 

Thank you for all this information, it will allow me to have a better understanding.

.config/gitpod/Dockerfile
```diff
@@ -12,11 +11,10 @@ ARG NPM_VERSION=9.6.7
 # Set a variable for the ungit version
 ARG UNGIT_VERSION=1.5.23
 
-# Update system and install packages, including Docker
-# Replace "python3.10" with "$(cut -d '=' -f 2 /tmp/python_version.config)" in the apt-get install command
+# Update system and install packages*
 # hadolint ignore=DL3008,DL3013
```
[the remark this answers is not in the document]
AurelienToitot: @ThomasSanson 

Indeed, since the packages have already been installed, there is no need to keep them in the image.

You will find my correction in this commit: 5d42bf05a1105e1186cc6a1eddf6f7b771451332

.config/gitpod/Dockerfile
AurelienToitot: @ThomasSanson 

I have just made the modification, and indeed the DL3008 and DL3009 errors for line 44 do not appear during testing.

You can find the modification in this commit: 59c3b6eeb5770fbcc4b5a9d946d47400702f0008

AurelienToitot: @ThomasSanson 


I have just made the modification, including the SHELL I had to place before hadolint. Otherwise, the modification for DL3008 and DL3009 errors would not be applied anymore.

You can find the modifications in this commit:7db31daba6df5509a1e118598c24e828218b5c3e

.config/gitpod/Dockerfile
AurelienToitot: I have just made the modification, and indeed the DL3008 and DL3009 errors for line 56 do not appear during testing.

You can find the modification in this commit: beaf0c0d0c14d522e0d782a413e38c4225387500

AurelienToitot: @ThomasSanson 

I have just made the modification related to the SHELL for line 56, and as with line 44, I had to introduce the SHELL command before hadolint.

You can find the modification in this commit: d56a4fbb2d1d46832ae5ee6929d7b3ba0d89660c

.config/gitpod/Dockerfile
AurelienToitot: @ThomasSanson 

I have just made the modification related to the SC2086 error, preventing globbing and word splitting. I added double quotes as recommended. However, when running the "make tests" command, the SC2086 error was still present. Therefore, I also added double quotes to the term "ungit@${UNGIT_VERSION}". After re-running the "make tests" command, no further errors were generated.

Nevertheless, I kindly request you to verify for security purposes.

You can find the modification in this commit: 94607ef9625e360875f22f7a766b3df8a2f71694

.config/gitpod/Dockerfile
AurelienToitot: @ThomasSanson 

I have just made the modification in accordance with the POSIX standard.

You can find the modifications in this commit: c88a85f8dbfef4874bbe07375307892dc3d75ae3

.config/gitpod/Dockerfile
AurelienToitot: @ThomasSanson 

I have just made the modification "FROM scratch" and after running "make tests" and "make docker-tests", I did not receive any error messages.

You can find the modifications in this commit: dbb88523c77c8201d1f7b5975d292e4f64d0eb68

.config/gitpod/Dockerfile
```diff
@@ -67,7 +67,7 @@ FROM npm-installer as cleanup
 RUN apt-get clean && rm -rf /var/lib/apt/lists/* tmp/* 
 
 # Final stage: Setup final image and user
-FROM ubuntu:jammy
+FROM scratch
```
vitabaks, line 70: Why?

Please use ubuntu or debian

AurelienToitot: This was a suggestion to avoid including the layers from the Ubuntu:jammy image in our final image. Since we are copying the root from the last stage instead of the Ubuntu:jammy root, I was leaning towards starting from the scratch image, as proposed by @thomas. However, if you prefer to stick with the Ubuntu:jammy image, that's not a problem. In that case, I can revert it back to Ubuntu.

vitabaks: @ThomasSanson please tell me what you think

ThomasSanson: I have tested the currently generated image in single stage mode, and it has a size of 903 MB with 14 layers.

In multi-stage mode, with an inheritance of ubuntu:jammy at the end, the image size is 903 MB with 9 layers.

By testing the final stage from scratch, the image size is 902 MB with only 3 layers.

In all cases, I have performed tests with "gp validate" and also tested the installation of Nginx just in case. I have also tested by forking @AurelienToitot's repository and merging it into the master branch (mine), and running `make tests`. Everything works perfectly fine.

Now that we have this data, here are my conclusions:

- Regardless of the final image, the size varies only slightly (1 MB), so it is not necessarily significant in terms of size.
- Starting from scratch significantly reduces the number of layers, which is significant. It is even a best practice, especially since it "hides" all the previous layers:
```bash
gitpod@thomassanso-postgresqlc-sejyfqwoj6r:/workspace/postgresql_cluster$ docker history postgresql_cluster:local
IMAGE          CREATED         CREATED BY                                      SIZE      COMMENT
bcfec7f0b666   4 minutes ago   /bin/sh -c #(nop)  USER gitpod                  0B        
06ec2516807f   4 minutes ago   /bin/sh -c useradd -l -u 33333 -G sudo -md /…   9.49kB    
87e04343e6fa   4 minutes ago   /bin/sh -c #(nop) COPY dir:807d72e12e5915c80…   825MB     
5a81c4b8502e   12 days ago     /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B        
<missing>      12 days ago     /bin/sh -c #(nop) ADD file:140fb5108b4a2861b…   77.8MB    
<missing>      12 days ago     /bin/sh -c #(nop)  LABEL org.opencontainers.…   0B        
<missing>      12 days ago     /bin/sh -c #(nop)  LABEL org.opencontainers.…   0B        
<missing>      12 days ago     /bin/sh -c #(nop)  ARG LAUNCHPAD_BUILD_ARCH     0B        
<missing>      12 days ago     /bin/sh -c #(nop)  ARG RELEASE                  0B     
```
vs
```bash
gitpod@thomassanso-postgresqlc-sejyfqwoj6r:/workspace/postgresql_cluster$ docker history postgresql_cluster:local
IMAGE          CREATED          CREATED BY                                      SIZE      COMMENT
de7b56120a18   12 minutes ago   /bin/sh -c #(nop)  USER gitpod                  0B        
339f6d780414   12 minutes ago   /bin/sh -c useradd -l -u 33333 -G sudo -md /…   9.49kB    
4088627bb2c6   12 minutes ago   /bin/sh -c #(nop) COPY dir:807d72e12e5915c80…   902MB 
```
This may seem a bit extreme for a simple development Dockerfile, but it fully complies with security best practices !

- Ultimately, it makes the image more stable and avoids potential sources of errors. If we need to change the environment and switch to a newer version of Ubuntu or completely change the distribution, we only need to modify one place instead of two. Therefore, starting from scratch promotes DRY (Don't Repeat Yourself) principles.

Just for information, whether it's Ubuntu, Debian, Fedora, etc., they all start from scratch, which is the base Docker image for building their operating systems. Alpine, on the other hand, starts from BusyBox.

Furthermore, since we copy the entire root to the root (`COPY --from=cleanup / /`), we overwrite everything. So, we stay within the guidelines.

With all this "expertise" (it's cool to dive deep ^^), from my side, I recommend the last layer to be "from scratch".

But honestly, it's a matter of perfectionism, so @vitabaks, as the project maintainer, I'll let you make the final decision.

.config/gitpod/Dockerfile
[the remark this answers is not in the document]
ThomasSanson: @AurelienToitot 

Due to your response, it appears possible that you may not have fully understood.

The DL3009 rule (DL indicating it comes directly from Hadolint) states:
Remove apt-get lists after installation.

In your multi-stage configuration, the purging of apt is not performed in the "builder" stage for Hadolint, which results in an error. However, you correctly perform the purge later in the "cleanup" stage.

Therefore, and simply because we are aware of the steps we have implemented, we can ignore this rule (DL3009) in the RUN command on line 17.

Indeed, we know that it is done later (the apt cache will be purged).

.config/gitpod/Dockerfile
ThomasSanson: Hello, it would be advisable to be cautious with copy-pasting, as it can lead to significant errors. 

Even though we are discussing in the comments, it's important not to develop bad reflexes.

In this particular case, it only pertains to rule DL3008, not DL3009 as mentioned in your comment. 

Nevertheless, the commit is fine 😇

.config/gitpod/Dockerfile
ThomasSanson: @AurelienToitot 

It's peculiar, in your commit there doesn't appear to be an added empty line at the end of the file.

Moreover, when I look at your branch, it's the same, no empty line at the end of the file.

.config/gitpod/Dockerfile
AurelienToitot: Thank you for the congratulations, it was a pleasure to contribute to this project.

Throughout the process of fixing and writing my code, I indeed used the commands `make tests` and `make docker-tests` to ensure that my modifications did not generate any other errors in the project during execution.

Regarding the `gp validate` command, I used it in conjunction with the `make reinitization` command to avoid keeping the same mount point during my tests.

Hoping to have been helpful to this project,

Best regards.

.config/gitpod/Dockerfile
vitabaks: thanks for the clarification

I'm not an expert in containers (since I don't use them in production), which is why the question arose.
Let's leave it as you suggest, I entrust you with this.

@AurelienToitot @ThomasSanson thank you for your contribution, it is important for the further development of the project.

PR #408, opened 2023-07-09
Haproxy stats to vip

chuegel:
This PR binds the `cluster_vip` IP to the haproxy stats endpoint if exists. 
It will align with the other exposed endpoints like master, replicas etc.
Please review.

Thanks

### DIFF
--- a/roles/haproxy/handlers/main.yml
+++ b/roles/haproxy/handlers/main.yml
@@ -1,7 +1,7 @@
 ---
 
 - name: Restart haproxy service
-  systemd:
+  ansible.builtin.systemd:
     daemon_reload: true
     name: haproxy
     enabled: true
@@ -9,9 +9,9 @@
   listen: "restart haproxy"
 
 - name: Check HAProxy is started and accepting connections
-  wait_for:
+  ansible.builtin.wait_for:
     port: "{{ haproxy_listen_port.stats }}"
-    host: "{{ hostvars[inventory_hostname]['inventory_hostname'] }}"
+    host: "{{ cluster_vip | default(hostvars[inventory_hostname]['inventory_hostname'], true) }}"
     state: started
     timeout: 120
     delay: 10
--- a/roles/haproxy/templates/haproxy.cfg.j2
+++ b/roles/haproxy/templates/haproxy.cfg.j2
@@ -21,11 +21,14 @@ defaults
 
 listen stats
     mode http
+{% if cluster_vip is defined and cluster_vip | length > 0 %}
+    bind {{ cluster_vip }}:{{ haproxy_listen_port.stats }}
+{% else %}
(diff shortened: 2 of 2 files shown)


roles/haproxy/templates/haproxy.cfg.j2
```diff
@@ -21,7 +21,11 @@ defaults
 
 listen stats
     mode http
+{% if cluster_vip is defined and cluster_vip | length > 0 %}
+    bind {{ cluster_vip }}:{{ haproxy_listen_port.stats }}
+{% else %}
     bind {{ hostvars[inventory_hostname]['inventory_hostname'] }}:{{ haproxy_listen_port.stats }}
+{% endif %}
```
vitabaks, line 28:
```suggestion
{% if cluster_vip is defined and cluster_vip | length > 0 %}
    bind {{ cluster_vip }}:{{ haproxy_listen_port.stats }}
{% endif %}
    bind {{ hostvars[inventory_hostname]['inventory_hostname'] }}:{{ haproxy_listen_port.stats }}
```

I suggest leaving listening to the local address, but also adding the possibility of listening to VIP

this will allow us to check the statistics for the host we need

roles/haproxy/handlers/main.yml
```diff
@@ -1,17 +1,17 @@
...
     enabled: true
     state: restarted
   listen: "restart haproxy"
 
 - name: Check HAProxy is started and accepting connections
-  wait_for:
+  ansible.builtin.wait_for:
     port: "{{ haproxy_listen_port.stats }}"
-    host: "{{ hostvars[inventory_hostname]['inventory_hostname'] }}"
+    host: "{{ cluster_vip | default(hostvars[inventory_hostname]['inventory_hostname'], true) }}"
```
vitabaks, line 14: I think this is an incorrect check because only one haproxy will be checked the one that is available by VIP

the initial task of this check was to check the status of each haproxy for each server in the balancers group

vitabaks: You can add a separate task that will check availability by VIP

vitabaks: "Check if HAProxy is accepting connections via the VIP address"

vitabaks: but this won't work because the keepalived role is executed later  

in this case, keepalived role must be performed before haproxy

chuegel: yes, you're right. But shouldn't the keepalived role be executed before haproxy anyways?

vitabaks: I don't have to, but I can do it.

chuegel: Well, in the end this PR is just cosmetics. So it would be also fine to leave it as it is.

vitabaks: Please close this PR if you do not plan to further refine this part of the code.

PR #409, opened 2023-07-09
make sure only pgbouncer can use this function

chuegel:
This PR follows the [best practice](https://www.pgbouncer.org/config.html) configuration to make sure only pgbouncer can use this function.

Please review
Thanks

### DIFF
--- a/roles/pgbouncer/config/tasks/main.yml
+++ b/roles/pgbouncer/config/tasks/main.yml
@@ -60,6 +60,8 @@
         SELECT usename, passwd FROM pg_shadow WHERE usename=$1;
         $$
         LANGUAGE sql SECURITY DEFINER;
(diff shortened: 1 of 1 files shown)


roles/pgbouncer/config/tasks/main.yml
```diff
@@ -60,6 +60,8 @@
         SELECT usename, passwd FROM pg_shadow WHERE usename=$1;
         $$
         LANGUAGE sql SECURITY DEFINER;
+        REVOKE ALL ON FUNCTION user_search(uname TEXT) FROM public;
+        GRANT EXECUTE ON FUNCTION user_search(uname TEXT) TO pgbouncer;
```
vitabaks, line 64:
```suggestion
        GRANT EXECUTE ON FUNCTION user_search(uname TEXT) TO {{ pgbouncer_auth_username }};
```

PR #413, opened 2023-07-13
Improve jinja spacing

pa-decarvalho:
This pull request aims to improve jinja spacing. To achieve this, the `jinja[spacing]` section has been removed from the `.config/ansible-lint.yml` file.

While making the correction, I came across the following `variable | bool` and decided to remove all `| bool` occurrences from the project as they are not necessary. Feel free to let me know if I should create another PR for this part or revert the change.

### DIFF
--- a/roles/consul/defaults/main.yml
+++ b/roles/consul/defaults/main.yml
@@ -4,11 +4,11 @@
 
 ## Core
 consul_debug: false
-is_virtualenv: "{{ lookup('env','VIRTUAL_ENV') | default('', true) }}"
+is_virtualenv: "{{ lookup('env', 'VIRTUAL_ENV') | default('', true) }}"
 consul_install_dependencies: true
 
 ### Package
-consul_version: "{{ lookup('env','CONSUL_VERSION') | default('1.8.7', true) }}"
+consul_version: "{{ lookup('env', 'CONSUL_VERSION') | default('1.8.7', true) }}"
 consul_architecture_map:
   # this first entry seems redundant
   # (but it's required for reasons)
@@ -29,9 +29,9 @@ consul_os: "\
   {% else %}\
     {{ ansible_system | lower }}\
   {% endif %}"
-consul_pkg: "consul{% if consul_enterprise %}-enterprise{% else %}{%endif%}_{{ consul_version }}_{{ consul_os }}_{{ consul_architecture }}.zip"
+consul_pkg: "consul{% if consul_enterprise %}-enterprise{% endif %}_{{ consul_version }}_{{ consul_os }}_{{ consul_architecture }}.zip"
 consul_zip_url: "https://releases.hashicorp.com/consul/{{ consul_version }}/consul_{{ consul_version }}_{{ consul_os }}_{{ consul_architecture }}.zip"
-consul_checksum_file_url: "https://releases.hashicorp.com/consul/{{ consul_version }}/consul_{{ consul_version}}_SHA256SUMS"
+consul_checksum_file_url: "https://releases.hashicorp.com/consul/{{ consul_version }}/consul_{{ consul_version }}_SHA256SUMS"
 
 ### Install Method
 consul_force_install: false
@@ -45,8 +45,8 @@ consul_config_template_path: "templates/config.json.j2"
 consul_configd_path: "/etc/consul.d"
 consul_bootstrap_state: "{{ consul_config_path }}/.consul_bootstrapped"
 consul_data_path: "/opt/consul"
-consul_log_path: "{{ lookup('env','CONSUL_LOG_PATH') | default('/var/log/consul', true) }}"
-consul_log_file: "{{ lookup('env','CONSUL_LOG_FILE') | default('consul.log', true) }}"
+consul_log_path: "{{ lookup('env', 'CONSUL_LOG_PATH') | default('/var/log/consul', true) }}"
+consul_log_file: "{{ lookup('env', 'CONSUL_LOG_FILE') | default('consul.log', true) }}"
 consul_run_path: "/run/consul"
 consul_binary: "{{ consul_bin_path }}/consul"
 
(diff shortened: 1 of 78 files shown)


vitabaks:
@pa-decarvalho

First and foremost, thank you for your contribution to the project. Your efforts towards improving jinja spacing are greatly appreciated, and your attention to detail is commendable.

However, I'd like to bring up the topic of removing all | bool occurrences. The use of the | bool filter in Ansible is considered a good practice. It ensures consistent boolean interpretation across various possible representations, such as "True", "true", "yes", "on", "1", etc. The | bool filter ensures these values are interpreted in their boolean representation, facilitating comparisons and usage in conditionals among others.

If you've encountered a situation where | bool is misused or not needed, that might indeed be a case for correction. But removing all occurrences of | bool without appropriate analysis and evaluation might lead to unforeseen consequences.

Looking forward to your thoughts on this.

cc @ThomasSanson

pa-decarvalho:
@vitabaks 

I didn't know that `| bool` was a good practice. Usually, when I work on an Ansible project, all the linters are active, and values other than `true` and `false` are not allowed for booleans (like `yes`, `1`...)

I will immediately replace all occurrences of `| bool` to avoid any unpleasant surprises.

vitabaks:
@ThomasSanson What do you think about `bool` filter?

pa-decarvalho:
> @ThomasSanson What do you think about `bool` filter?

@vitabaks @ThomasSanson 

I am in the process of reverting back regarding the `bool` filter and correcting only lint errors (I am waiting for the tests to pass locally).

If you agree, I can create an issue for us to discuss the pros and cons of using it.

vitabaks:
thread for discussion of the `bool` filter
https://github.com/vitabaks/postgresql_cluster/discussions/415

ThomasSanson (approved):
@vitabaks 

I have completed the task of reviewing all 28 files.

I can confirm that everything is in order from my perspective.

Great job, @pa-decarvalho, and thank you! 🎉🙏

Now, let's move on to the next task! 💪😊

roles/consul/defaults/main.yml
```diff
@@ -29,9 +29,9 @@ consul_os: "\
   {% else %}\
     {{ ansible_system | lower }}\
   {% endif %}"
-consul_pkg: "consul{% if consul_enterprise %}-enterprise{% else %}{%endif%}_{{ consul_version }}_{{ consul_os }}_{{ consul_architecture }}.zip"
+consul_pkg: "consul{% if consul_enterprise %}-enterprise{% endif %}_{{ consul_version }}_{{ consul_os }}_{{ consul_architecture }}.zip"
```
vitabaks, line 32: it looks like the "else" has been removed here. Why?

image

roles/consul/defaults/main.yml
```diff
@@ -61,30 +61,31 @@ consul_systemd_limit_nofile: 65536
...
-consul_iface: "\
-  {% if ansible_os_family == 'Windows' %}\
-    {{ lookup('env','CONSUL_IFACE') | default(ansible_interfaces[0].interface_name, true) }}\
-  {% else %}\
-    {{ lookup('env','CONSUL_IFACE') | default(ansible_default_ipv4.interface, true) }}\
-  {% endif %}"
-consul_node_role: "{{ lookup('env','CONSUL_NODE_ROLE') | default('client', true) }}"
-consul_recursors: "{{ lookup('env','CONSUL_RECURSORS') | default('[]', true) }}"
-consul_bootstrap_expect: "{{ lookup('env','CONSUL_BOOTSTRAP_EXPECT') | default(false, true) }}"
+consul_iface: >
```
vitabaks, line 79: it looks like there was a small refactoring of the code here, although this is not in the description of the PR

testing of the 'dcs_type: consul' schema has not been automated yet, has there been an attempt to deploy a cluster of Type C schema manually?

roles/consul/defaults/main.yml
```diff
@@ -122,21 +123,22 @@ consul_cloud_autodiscovery_params: ""
...
 ### Addresses
-consul_bind_address: "\
-  {% if ansible_system == 'FreeBSD' or ansible_system == 'Darwin' %}\
-    {{ lookup('env','CONSUL_BIND_ADDRESS') | default(hostvars[inventory_hostname]['ansible_'+ consul_iface ]['ipv4'][0]['address'], true) }}\
-  {% elif ansible_os_family == 'Windows'  %}\
-    {{ lookup('env','CONSUL_BIND_ADDRESS') | default(hostvars[inventory_hostname]['ansible_ip_addresses'][0], true) }}\
-  {% else %}\
-    {{ lookup('env','CONSUL_BIND_ADDRESS') | default(hostvars[inventory_hostname]['ansible_'+ consul_iface | replace('-', '_')]['ipv4']['address'], true) }}\
-  {% endif %}"
+consul_bind_address: >
```
vitabaks, line 126: Added `consul_bind_env`

roles/consul/defaults/main.yml
ThomasSanson: The {% else %} part was doing nothing, so it is unnecessary.

We can see that if the `consul_enterprise` condition is met, it adds "-enterprise," and that's all.

Therefore, if the condition is fulfilled, we will have:
consul-enterprise_{{ consul_version }}

If it is not fulfilled, we will have:
consul_{{ consul_version }}

In my opinion, this refactoring is justified

roles/consul/defaults/main.yml
ThomasSanson: Yes, I had seen this optimization.

Based on my analysis, it appears to be correct and more readable.

However, it is indeed outside the scope of this PR. Could it be considered premature optimization?

In any case, it is true that we sorely lack tests for this particular part (Type C).

I will try to see if I can come up with an Ansible test code for this part (before and after the change) to verify the expected results.

roles/consul/tasks/services.yml
```diff
@@ -40,15 +40,14 @@
 
 - name: Set fact with list of existing configuration files
   ansible.builtin.set_fact:
-    list_current_service_config: "{{ list_current_service_config |default([]) + [ config_file_item.path ] }}"
+    list_current_service_config: "{{ list_current_service_config | default([]) + [config_file_item.path] }}"
```
vitabaks, line 43: > [config_file_item.path] 

why have the spaces been removed here on the contrary?

roles/consul/defaults/main.yml
vitabaks: ok, thanks

roles/consul/defaults/main.yml
vitabaks: > However, it is indeed outside the scope of this PR. Could it be considered premature optimization ?

I am not against refactoring it is for the benefit of the project, but it is advisable to describe all the changes in the description of PR or create separate PR for different tasks, for example jinja spacing is one thing and refactoring is another PR.

> In any case, it is true that we sorely lack tests for this particular part (Type C).

I have already allocated several hours, but it has not worked out yet, there are difficulties in the fact that the dcs_type value cannot be redefined at the level of a separate molecule test since 'vars_files' is used for vars/main.yml

If you have any ideas, I will be grateful. But this is beyond the scope of the topic of this PR, you can discuss in the issue https://github.com/vitabaks/postgresql_cluster/issues/375

PR #438, opened 2023-08-10
Add ansible_host with fallback on etcd,pg_hba, haproxy conf

tanganellilore:
Hi @vitabaks,

This PR introduce the usage of `ansible_host` that can be used in inventory to specify IP of machine, instead usage of `inventory_hostname`.

I notice in the invnetory example that this was specified, but hostnamme (if we use it on inventory) not working whell in etcd, pg_hba and haproxy.

let me know if i need to change something.
Thanks


vitabaks:
Hi @tanganellilore 

ansible_host is used to specify a public IP address to connect to servers during deployment. 

Therefore, using this variable for listening by cluster components is a bad idea.

vitabaks:
isn't this comment enough for the user to specify the ip address and not the names?

> # Please specify the ip addresses and connection settings for your environment
> # The specified ip addresses will be used to listen by the cluster components.
> # Attention! Specify private IP addresses so that the cluster does not listen a public IP addresses.
> # For deploying via public IPs, add 'ansible_host=public_ip_address' variable for each node.

tanganellilore:
Hi @vitabaks,
Yes ansible_host is used to connect into machine, but this not mean that is a "public" ip. I can have a private network, with private AD and with him hostname.

In perspective to use this very usefull playbook, in AWX or APP, usage of ip as inventory hostname for me is a limitation. 

In any case, can you evaluate to use a different variable for ip, instead `inventory_hostname`, that is not ansible_hosts? 
I can implement it for you. 

One idea can be the usage of default_ipv4 facts (but i don't know if in Aws like is the private ip) or pass the ifname to retrieve current ip. 

Let me know and thanks for support

vitabaks:
using `inventory_hostname` was a simple and fast way to implement listening settings by cluster components on the specified network. 

It may be worth abandoning this method in favor of the `bind_address` variable (similar to `consul_bind_address`) for the interface designated in the `interface` variable (similar to `vip_interface` of `consul_iface`)

tanganellilore:
Ok, il will check it and propose in this PR this approach.
I will think to maintain the fallback to inventory_hostname as per default.

It's ok for you?

vitabaks:
I don't mind as long as it's good for everyone

PR #439, opened 2023-08-17
Added the functionality of assigning roles to users and few improvements

sgremyachikh:
1. I would like to see the results of the deployment with details also in case of a cluster configuration change
2. I would like to see a list of users after deploying the database with the user roles, if they are specified
3. Added the functionality of assigning roles to users
4. I added an example in postgresql_cluster/vars/main.yml : we are adding the pg_monitor system role to the monitoring user


roles/deploy-finish/tasks/main.yml
```diff
@@ -59,6 +59,24 @@
...
+      register: users_result
+      delegate_to: "{{ groups.master[0] }}"
+      changed_when: false
+
+    - name: PostgreSQL list of users
+      run_once: true
+      ansible.builtin.debug:
+        var: users_result.stdout_lines
+  ignore_errors: true
+  tags: databases, db_list, cluster_info, cluster_status, point_in_time_recovery
```
vitabaks, line 78:
```suggestion
  tags: users, users_list, cluster_info, cluster_status, point_in_time_recovery
```

roles/postgresql-users/tasks/main.yml
```diff
@@ -21,4 +21,20 @@
   when: postgresql_users is defined and postgresql_users | length > 0
   tags: postgresql_users
 
+- name: Grant roles to users
+  community.postgresql.postgresql_membership:
+    group: "{{ item.role }}"
```
vitabaks, line 26:
```suggestion
    group: "{{ item.role | default('') }}"
```
If the 'role' is not specified in postgresql_users, the task will fail because it will try to get a non-existent attribute.

roles/postgresql-users/tasks/main.yml
```diff
@@ -21,4 +21,20 @@
...
+    login_host: "127.0.0.1"
+    login_port: "{{ postgresql_port }}"
+    login_user: "{{ patroni_superuser_username }}"
+    login_password: "{{ patroni_superuser_password }}"
+    state: present
+  ignore_errors: true
+  loop: "{{ postgresql_users | flatten(1) }}"
+  loop_control:
+    label: "{{ item.name }}"
+  when: postgresql_users is defined and postgresql_users | length > 0
```
vitabaks, line 37:
```suggestion
  when:
    - postgresql_users is defined
    - postgresql_users | length > 0
    - item.role | default('') | length > 0
```

If the 'role' is defined, but empty, then community.postgresql.postgresql_membership may not work correctly or give an error, since an empty value for the role may be invalid.

vitabaks:
Thanks @sgremyachikh

sgremyachikh:
Thank you for your job, it’s awesome solution to deploy ha Postgres cluster

PR #450, opened 2023-08-23
issues 449,448 remove_cluster.yml improvements

rrrru:
@vitabaks vars -> default(false), blocks for dry, remove systemd file cuz consul role do not restart service if exist
https://github.com/vitabaks/postgresql_cluster/issues/449

### DIFF
--- a/remove_cluster.yml
+++ b/remove_cluster.yml
@@ -7,28 +7,29 @@
     - vars/{{ ansible_os_family }}.yml
   tasks:
     - block:
-      - name: Stop Patroni service
-        ansible.builtin.service:
-          name: patroni
-          state: stopped
-          enabled: false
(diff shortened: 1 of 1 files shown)


rrrru:
@vitabaks I set 'ignore_errors' because the task crashes when restarted.

vitabaks:
@rrrru thanks!

I see that you decided not to add the task 'consul kv delete', why?

rrrru:
> @rrrru thanks!
> 
> I see that you decided not to add the task 'consul kv delete', why?

At first I had wrong idea what the problem was.
It turned out that the problem was in the scope to which the task was distributed (postgre_cluster). So I added another group for conusl and specified consul_instances in hosts.

PR #475, opened 2023-09-26
confd requires an additional 'basic_auth' flag for etcd authorization

lavr:
During detailed testing, it was found that the 'username' and 'password' parameters alone are insufficient for the proper operation of confd with etcd authentication. It is necessary to add the 'basic-auth' flag.


vitabaks:
Thanks @lavr !

PR #476, opened 2023-09-27
Patroni pre-checks: check etcd cluster accessible

lavr:
I suggest checking the network availability of the etcd cluster before adding nodes to the Patroni cluster. If etcd is unavailable for some reason, don't even attempt to add the node.

I created the role patroni/pre-checks for this purpose and added it to the playbooks deploy_pgcluster.yml and add_pgnode.yml. 

I considered adding this checks to other playbooks but decided it might hinder the administrator from updating the cluster in some situations, which could be unsafe.

### DIFF
--- a/add_pgnode.yml
+++ b/add_pgnode.yml
@@ -46,6 +46,12 @@
         var: new_nodes
       tags: always
 
+    - name: "Patroni pre-checks"
+      ansible.builtin.include_role:
+        name: patroni/pre-checks
+      when:
+        - new_node | default(false) | bool
+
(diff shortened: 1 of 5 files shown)


vitabaks:
@lavr Thanks!

Maybe it 's worth move this task to an already existing pre-checks role?

- https://github.com/vitabaks/postgresql_cluster/tree/master/roles/pre-checks/tasks

the playbook config_pgcluster.yml has a variable `postgresql_cluster_maintenance` by which you can use which you can limit the execution of tasks only for this playbook

- https://github.com/vitabaks/postgresql_cluster/blob/master/config_pgcluster.yml#L12

PR #483, opened 2023-10-05
Change to avoid the error "AnsibleUndefinedVariable: 'ansible_hostname'"

TommasoDb:
This commit fixes the error "AnsibleUndefinedVariable: 'ansible_hostname'". This error only occurs when the playbook `balancers.yml` is executed directly using the command
```
ansible-playbook balancers.yml
```
Host `postgres_cluster` is not part of this playbook, but their facts are needed in the template haproxy.cfg.j2

### DIFF
--- a/roles/haproxy/tasks/main.yml
+++ b/roles/haproxy/tasks/main.yml
@@ -5,6 +5,7 @@
   delegate_to: "{{ item }}"
   delegate_facts: true
   loop: "{{ groups['postgres_cluster'] }}"
+  when: hostvars[groups['postgres_cluster'][0]].ansible_hostname is not defined
 
 # Install HAProxy from rpm/deb packages
 # from repo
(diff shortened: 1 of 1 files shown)


vitabaks:
ok, thanks @TommasoDb

roles/haproxy/tasks/main.yml
```diff
@@ -1,5 +1,11 @@
 ---
 
+- name: Gather facts for hosts postres_cluster
+  ansible.builtin.setup:
+  delegate_to: "{{ item }}"
+  delegate_facts: true
+  loop: "{{ groups['postgres_cluster'] }}"
```
vitabaks, line 7:
```suggestion
- name: Gather facts for hosts postres_cluster
  ansible.builtin.setup:
  delegate_to: "{{ item }}"
  delegate_facts: true
  loop: "{{ groups['postgres_cluster'] }}"
  when: hostvars[groups['postgres_cluster'][0]].ansible_hostname is not defined
```

perform the task only when there are no facts

PR #491, opened 2023-10-25
roles: pgbouncer: drop `-q` option from call

jimnydev:
Drop unnecessary  `q`  option

### DIFF
--- a/roles/pgbouncer/config/tasks/main.yml
+++ b/roles/pgbouncer/config/tasks/main.yml
@@ -26,7 +26,7 @@
       become: true
       become_user: postgres
       ansible.builtin.command: >-
-        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXcq
+        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXc
         "SELECT concat('\"', usename, '\" \"', passwd, '\"') FROM pg_shadow where usename != '{{ patroni_replication_username }}'"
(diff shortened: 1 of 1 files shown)


roles/pgbouncer/config/tasks/main.yml
```diff
@@ -26,7 +26,7 @@
       become: true
       become_user: postgres
       ansible.builtin.command: >-
-        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXcq
+        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXqc
```
vitabaks, line 29:
```suggestion
        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXc
```

I think the `-q `option is not required when using -`tAXc`

kornrunner: I haven't checked in detail, just tried to make the change least disruptive. 

Nevertheless, I dropped the `-q` option in following commit. Thanks!

jimnydev: I haven't checked in detail, just tried to make the change least disruptive.

Anyhow, I've dropped the `-q` option in following update. Thanks!

vitabaks: @jimnydev Thanks. Please test it.

jimnydev: It took me a bit, but I did test both changes (swapping the options and dropping `q) - both are working fine. Thank you for merging it!

PR #530, opened 2023-12-16
Adding variable and corresponding code for diverting percentage of requests which coming to a slave towards the master

vreitech:
The pull request adding variable and corresponding code for diverting specified percentage of requests which coming into HAProxy to a slave port towards the master.

Necessity of the feature is questionable and discussionable. AFAIK there's no any direct profit to use routing of a part of reading queries to master instead of a slave.

**Changes**

* added `haproxy_divert_to_master` var into `vars/main.yml`

* setting `haproxy_divert_to_master` to 0 to avoid changes in ansible deployment result with default configuration (which disabling any diverting of connections to master instead of slave)

* added corresponding code into `haproxy.cfg.j2` template

* added corresponding code into `haproxy.tmpl.j2` template

### DIFF
--- a/roles/confd/templates/haproxy.tmpl.j2
+++ b/roles/confd/templates/haproxy.tmpl.j2
@@ -25,13 +25,7 @@ listen stats
     stats enable
     stats uri /
 
-listen master
-{% if cluster_vip is defined and cluster_vip | length > 0 %}
-    bind {{ cluster_vip }}:{{ haproxy_listen_port.master }}
-{% else %}
-    bind {{ hostvars[inventory_hostname]['inventory_hostname'] }}:{{ haproxy_listen_port.master }}
-{% endif %}
-    maxconn {{ haproxy_maxconn.master }}
+backend master_backend
     option tcplog
     option httpchk OPTIONS /primary
     http-check expect status 200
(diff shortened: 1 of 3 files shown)


vitabaks:
> Necessity of the feature is questionable and discussionable. AFAIK there's no any direct profit to use routing of a part of reading queries to master instead of a slave.

@vreitech Thank you for submitting this PR. The idea of diverting a certain percentage of read queries from HAProxy to the master node is interesting, but its practical utility seems questionable. Generally, the purpose of directing traffic to a slave is to distribute the load and avoid overburdening the master with read queries, especially in high-traffic environments.

Your implementation suggests that a portion of read queries can be intentionally rerouted to the master. This might defy the typical use case of a read replica, which is to offload read operations from the master. Could you please elaborate on specific scenarios or use cases where this would be beneficial?

Additionally, if there's a need to direct more traffic to the master, wouldn't adjusting the existing load balancing setup or adding a new endpoint like '/read-only' to point to the master be a more straightforward solution? This approach could offer more transparency and control over traffic routing without the need for a new variable or complex configuration changes.

```
GET /read-only: like the above endpoint, but also includes the primary.
GET /read-only-sync: like the above endpoint, but also includes the primary.
```
https://patroni.readthedocs.io/en/latest/rest_api.html#health-check-endpoints

PR #562, opened 2024-01-27
Add auto_compaction for maintenance etcd on etcd.conf

SDV109:
This PR is designed to automate the maintenance of etcd cluster. Namely, the addition of automatic compression of the etcd database.

> Since etcd keeps an exact history of its keyspace, this history should be periodically compacted to avoid performance degradation and eventual storage space exhaustion.

If you are using this change with an existing etcd cluster, after the configuration change and the automatic compression period has passed, it is recommended to perform a one-time defragmentation. 

It is described in more detail here: https://etcd.io/docs/v3.5/op-guide/maintenance/


vitabaks:
@SDV109 Thank you for contribution!

vitabaks (approved):

PR #573, opened 2024-02-06
Python 3.11 use on RedHat 8 and above 

weisscorp:
Hi there!

We've encountered the need to upgrade Python to version 3.11 in our company. I've made some minor code adjustments to ensure compatibility with Python 3.11 by default. Since we don't have Debian servers, I wasn't able to test this. The primary constraint for using Python 3.11 is RedHat 8 and above. Kindly review my commit.

Additionally:

- Correct variable formatting for TimescaleDB installation task (f91d4ea4b209987f5c0fd3e4c83d20360c501f15) to improve the reliability of the code.

### DIFF
--- a/roles/packages/tasks/main.yml
+++ b/roles/packages/tasks/main.yml
@@ -159,12 +159,12 @@
 # PostgreSQL prepare for install (for RHEL)
 - block:
     - name: PostgreSQL | check if appstream module is enabled
-      ansible.builtin.command: '/usr/bin/dnf -y -C module list postgresql'
+      ansible.builtin.command: 'dnf -y -C module list postgresql'
       register: postgresql_module_result
       changed_when: false
 
     - name: PostgreSQL | disable appstream module
-      ansible.builtin.command: '/usr/bin/dnf -y -C module disable postgresql'
+      ansible.builtin.command: 'dnf -y -C module disable postgresql'
       when: "'[x] ' not in postgresql_module_result.stdout"
   when: installation_method == "repo" and
         ansible_os_family == "RedHat" and


vitabaks:
@weisscorp Thank you for your contribution.

Could you describe in more detail what is the reason for the need to pin the Python version to 3.11? Why should this be redefined at the postgresql_cluster project level (and therefore for everyone)? 

I'm clarifying because variables can (and some need to) be redefined to suit your requirements, it doesn't always need to be changed at the project level.

weisscorp:
In fact, the request may not seem significant, but primarily for security reasons. Patroni in the API gives the version of Python and the version of Patroni itself. Secondly, in order to maintain the current version of Python, versions 3.6 and 3.7 are no longer supported. And security patches are released for 3.8-3.10.

vitabaks:
@weisscorp I made a minor correction 5a2c6ac5c6c121ad4e391037e1c56371e8932bfd

vitabaks:
@weisscorp if you have time, try to figure out the cause of the `No module named 'dnf` error on RHEL 9.

I understand that here we need the full path to `dnf` tool, or adding the `PATH` variable to the tasks, but it is not clear why this was needed now with Python 3.11

weisscorp:
@vitabaks Sure, I'll take a look

weisscorp:
I couldn't reproduce the problem in the image, so I specified the absolute path. It's worth trying to run the tests again

weisscorp:
What is this for? I think there is no need to touch the system DNS servers. They are in no way related to patroni.

`Add DNS server(s) into /etc/resolv.conf`

weisscorp:
just try restarting the tests, it seems it was a momentary problem related to the update. To avoid this, you should add a retry

vitabaks:
> To avoid this, you should add a retry

Feel free to suggest a new PR.

weisscorp:
I don’t like this error, as if the default Python for the entire system had changed to a new version. It was necessary to change only for patroni. Give me some time so I can review the PR and code

weisscorp:
@vitabaks  I'm not sure exactly what container you're using, but the default [oraclelinux:9] already written correctly.

`#!/usr/bin/python3.9`

vitabaks:
> I'm not sure exactly what container you're using

`glillico/docker-centosstream9-ansible:latest`

Issue #582, opened 2024-02-21
Can't connect to db, using SSL certificate over pgbouncer.


davideferrero:
Hi everyone!

We just finished configuring the PostgreSQL cluster, and everything looks great! We have set up the SSL connection for the cluster. However, when we try to connect to the VIP at port 5000, we go through pgbouncer, and it doesn't allow SSL connections. Instead, if we connect directly to the node at port 5432, we establish the SSL connection without any issues.

`❯ PGSSLMODE=require psql --host pgdebian2.test.it.lan --port 5000 --user testuser --db test --password
Password:
psql: error: connection to server at "pgdebian2.test.it.lan" (192.168.43.51), port 5000 failed: server does not support SSL, but SSL was required`

Thanks a lot! :)

n-borges:
Hey Davide! 

you can add the following lines to `/etc/pgbouncer/pgbouncer.ini` to enable tls:

```
client_tls_sslmode = allow
client_tls_key_file = <key file path>
client_tls_cert_file = <cert file path>
```

It would be nice to add these configurations via playbook variables, i wonder if this is supported

vitabaks:
Feel free to create a PR.

n-borges:
> Feel free to create a PR.

I will happily work on it in the next few days

PR #585, opened 2024-02-22
feat: adds variables to enable client tls on pgbouncer

n-borges:
This PR is related to #582.

Although this is a trivial change, I think that quite a few people could incur in the requirement of having to enable TLS at least between clients and pgbouncer.

This has been tried on a deployment of type "A". Without these configurations, trying to connect to the cluster `vip` using TLS fails:

```bash
❯ PGSSLMODE=require psql --host pgdebian2.domain.lan --port 5000 --user testuser --db test --password
psql: error: connection to server at "pgdebian2.domain.lan" (192.168.99.99), port 5000 failed: server does not support SSL, but SSL was required
```

To allow encrypted connections, one could add to `/etc/pgbouncer/pgbouncer.ini` on each node:

```
# using the self-signed example certs
client_tls_sslmode = allow
client_tls_key_file = /etc/ssl/private/ssl-cert-snakeoil.key 
client_tls_cert_file = /etc/ssl/certs/ssl-cert-snakeoil.pem
```

After restarting the `pgbouncer` service, encrypted connections are allowed on the cluster virtual ip.

```bash
❯ PGSSLMODE=require psql --host pgdebian2.domain.lan --port 5000 --user testuser --db test --password
psql (16.1, server 16.2 (Debian 16.2-1.pgdg120+2))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, compression: off)
```

Let me know if you want me to add some tests.

Thanks!

### DIFF
--- a/roles/pgbouncer/templates/pgbouncer.ini.j2
+++ b/roles/pgbouncer/templates/pgbouncer.ini.j2
@@ -36,6 +36,10 @@ listen_backlog = 4096
 max_prepared_statements = {{ pgbouncer_max_prepared_statements }}
 so_reuseport = 1
 
+client_tls_sslmode = {{ pgbouncer_client_tls_sslmode }}
+client_tls_key_file = {{ pgbouncer_client_tls_key_file }}
+client_tls_cert_file = {{ pgbouncer_client_tls_cert_file }}
+
 log_connections = 0
 log_disconnections = 0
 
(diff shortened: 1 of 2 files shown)


vitabaks:
Hi! Thank you for your contribution to the project.

Could you add an example of a setup with the test result to the PR? In which the encrypted connection would be visible?

n-borges:
> Hi! Thank you for your contribution to the project.
> 
> Could you add an example of a setup with the test result to the PR? In which the encrypted connection would be visible?

Of course. I updated the PR body

PR #591, opened 2024-02-29
Using deb822_repository module instead of apt_key and apt_repository

FactorT:
Since `apt_key` is deprecated in Debian12 and Ubuntu 24.04, the `deb822_repository` module must be used instead.

Major changes:
- new tasks "Add repository and repository apt-key" for installation:
  - postgres
  - consul
  - pg_probackup
  - pg_backrest

Additional minor changes:
- new path for postgresql.org repo key (key the same)
- new key for pg_probackup (old key doesn't work)


FactorT:
I saw the problem with tests: role "add-repository" demand python3-debian package and is executed before role "packages" which install python3-package

vitabaks:
Hello @FactorT 

I don't quite understand if this change is really necessary, since all tests for Debian 12 pass successfully (every day) - https://github.com/vitabaks/postgresql_cluster/actions/workflows/schedule_pg_debian12.yml

FactorT:
@vitabaks 
Yes this is not really necessary right now. Because apt-key is just marked as depricated but still works in Debian12.
Probably apt-key will not work in next major release. And you can reject or hold this PR and come back to it during prepare to next major release of Debian.
More information at this [medium](https://medium.com/@glegoux/ubuntu-22-04-jammy-jellyfish-apt-key-is-deprecated-2dbbee8aec84) or [Geerling](https://www.jeffgeerling.com/blog/2022/aptkey-deprecated-debianubuntu-how-fix-ansible) or [page](https://sites.google.com/site/installationubuntu/home/ubuntu-22-04/apt-key-is-deprecated)

vitabaks:
Thank you for the information.

vitabaks:
> apt-key is deprecated from Ubuntu 22.04 LTS (Jammy Jellyfish)

It seems that this does not only apply to Debian.

> Ansible now has the [ansible.builtin.deb822_repository](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/deb822_repository_module.html) module, which can add keys and repositories in one task. It's a little more complex than the old way, and requires Ansible 2.15 or later

this must be taken into account and the `minimal_ansible_version`, variables `min_ansible_version` and README  must be updated.

I also think it's worth checking if the '`ansible.built in.deb822_repository`' module is  with oldest versions of Debian and Ubuntu and if so, then just replace the 'ansible.builtin.apt_key' and 'ansible.builtin.apt_repository' modules.

vitabaks:
I also remembered that in one of the projects I used [keyring](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html#parameter-keyring) for `apt_key` module.

Example:

```yaml
    - name: Add repository apt-key
      ansible.builtin.apt_key:
        url: "{{ item.key }}"
        state: present
        keyring: /etc/apt/trusted.gpg.d/{{ item.key | urlsplit('netloc') }}.gpg
      loop: "{{ apt_repository_keys }}"
      register: apt_key_status
      until: apt_key_status is success
      delay: 10
      retries: 3
      when: apt_repository_keys | length > 0
```

so it's worth considering whether this is the best alternative to the new module.

FactorT:
> I also remembered that in one of the projects I used [keyring](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html#parameter-keyring) option for `apt_key` module.
>
> so it's worth considering whether this is the best alternative to the new module (it would require fewer changes).

Yes, it's interesting idea. But keyring option uses [apt-key bin file](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/apt_key.py#L372)

FactorT:
I don't understand what's wrong with debin12 and debian11 molecule's tests.
In my local environment config_pgcluster playbook works fine.
Module python3-debian is presented at molecule's host:
[string 457](https://github.com/vitabaks/postgresql_cluster/actions/runs/8109426897/job/22164585561?pr=591#step:6:463) python3-debian was installed
[string 462](https://github.com/vitabaks/postgresql_cluster/actions/runs/8109426897/job/22164585561?pr=591#step:6:468) add repository task was done succesfully
It was role add-repository in deploy_pgcluster playbook
Next playbook config_pgcluster uses the same role add-repository which shows that python3-debian package is installed at [string 1233](https://github.com/vitabaks/postgresql_cluster/actions/runs/8109426897/job/22164585561?pr=591#step:6:1239)
But next task add-repository can't find this module.

vitabaks:
OK, I'll try to figure it out.

roles/consul/tasks/install_linux_repo.yml
```diff
@@ -61,18 +61,37 @@
...
-    - name: Add Debian/Ubuntu Linux repository
-      ansible.builtin.apt_repository:
-        repo: "deb {{ consul_repo_url }} {{ ansible_distribution_release }} main"
+      register: apt_status
+      until: apt_status is success
+      delay: 5
+      retries: 3
+      when: ansible_os_family == "Debian"
+
+    - name: Add Debian/Ubuntu Linux repository and signing key, uses whichever key is at the URL
```
vitabaks, line 75:
```suggestion
    - name: Add Debian/Ubuntu Linux repository and signing key
```

vitabaks:
@FactorT I have made some changes to the code, please check it out. But I have not yet found a solution to the error in debian 11, 12 when the second execution of the playbook.

FactorT:
@vitabaks thank you very much!

FactorT:
@vitabaks How did you fix the error in debian 11, 12 when the second execution of the playbook?
btw, I changed consul's installation from apt to deb822

vitabaks:
Hi @FactorT 

> How did you fix the error in debian 11, 12 when the second execution of the playbook?

I don't know ) maybe the ansible update helped.

vitabaks:
Another theory is that it is possible that fixes have been made to system packages such as python3 that contribute to the correct execution of the code now.

vitabaks:
I've done a few additional test runs and it looks like everything is fine now.

I will combine this PR and if there is a problem in the future, we will promptly fix them.

PR #647, opened 2024-05-02
Add basic security for Patroni REST API

rcknr:
This PR adds two variables: `patroni_restapi_username` and `patroni_restapi_password` to add authentication for Patroni REST API providing basic security for unsafe operations.

### DIFF
--- a/vars/main.yml
+++ b/vars/main.yml
@@ -18,6 +18,8 @@ patroni_superuser_username: "postgres"
 patroni_superuser_password: "postgres-pass"  # please change password
 patroni_replication_username: "replicator"
 patroni_replication_password: "replicator-pass"  # please change password
+patroni_restapi_username: "patroni"
+patroni_restapi_password: "restapi-pass"  # please change password
 
 synchronous_mode: false  # or 'true' for enable synchronous database replication
 synchronous_mode_strict: false  # if 'true' then block all client writes to the master, when a synchronous replica is not available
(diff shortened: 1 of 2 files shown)


vars/main.yml
```diff
@@ -18,6 +18,8 @@ patroni_superuser_username: "postgres"
 patroni_superuser_password: "postgres-pass"  # please change password
 patroni_replication_username: "replicator"
 patroni_replication_password: "replicator-pass"  # please change password
+patroni_restapi_username: "patroni"
+patroni_restapi_password: "restapi-pass"  # please change password
```
vitabaks, line 22: I suggest placing these variables under the `patroni_restapi_port` variable to group the restapi variables.

vitabaks:
Thanks @rcknr

vitabaks:
Please think about backward compatibility. 

What happens if we update the code and use it to cluster the deployed previous version of the code?

What happens if new variables are not specified or empty values? 

What happens if variables with values are specified and the code is executed in a cluster in which authorization has not yet been enable.

rcknr:
The way Ansible URI module works it's going to send an unauthenticated request first and upon receiving a 401 code (which Patroni does) will repeat the request with provided username and password. This behaviour is described under [`force_basic_auth`](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/uri_module.html#parameter-force_basic_auth) key which is forcing immediate authentication. This address the case when variables are set but deployed cluster is configured differently (your first and last points).
Regarding the middle point, it makes sense to add the filter.

vitabaks:
@rcknr please look in the code for other places where interaction with the REST API is performed, I'm sure there were more such tasks

Try to search by `url: http://{{ inventory_hostname }}:{{ patroni_restapi_port }}`

rcknr:
Yes, there are more but most of them are `GET` operations which don't require authentication in Patroni.

Issue #612, opened 2024-03-28
Using IP address instead of hostname in Ansible AWX/Tower #question


silverbullet0201:
Current situation:
I use ansible tower (a.k.a. AWX) to manage our infrastructure
Names of our hosts are human-readable names like `pg-node-01`, `pg-node-02`, ...

When I run `deploy_pgcluster.yaml` playbook it fails in `Enable and start etcd service` task because in the etcd settings like `ETCD_LISTEN_CLIENT_URLS` and `ETCD_ADVERTISE_CLIENT_URLS` hostname (value from `inventory_hostname`) is used instead of host's IP address

---

My question is next:
Can we make playbook roles to use IP-address of the hostname (from `ansible_host` variable from inventory for example) instead of inventory hostname (`inventory_hostname` variable is in use currently)

Maybe there's option like adding variable `ansible_host` to inventory and replacing `{{ inventory_hostname }}` to `{{ ansible_host | default(inventory_hostname) }}` in `./roles/etcd/templates/etcd.conf.j2` (or in any other place that required)

---

There's some example:
IP-address is used instead of hostname (etcd service is running and working)
```bash
ETCD_LISTEN_CLIENT_URLS="http://1.2.3.4:2379,http://127.0.0.1:2379"
ETCD_ADVERTISE_CLIENT_URLS="http://1.2.3.4:2379"
# ...
```

hostname is used instead of IP-address (etcd service is stopping just after start)
``` bash
ETCD_LISTEN_CLIENT_URLS="http://pg-node-01:2379,http://127.0.0.1:2379"
ETCD_ADVERTISE_CLIENT_URLS="http://pg-node-01:2379"
# ...
```

vitabaks:
Yes, it must be done. It is not the first request on the topic inventory_hostname.

PR #649, opened 2024-05-05
Add small pg_cluster_default role. Add ability to define custom runni…

dtrdnk:
Add ability use specific IP. Fix inventory group etcd_cluster: use hostnames instead of ip-address #612


vitabaks:
Thanks for this PR! 

Could you help me understand a bit more about why you've classified this as a fix? It would be great if you could describe the specific issue this PR addresses. 

Also, I'm curious about the decision to create a separate role. Could you share your thoughts on why this approach was chosen over defining variables in `vars/main.yml`?

dtrdnk:
> Could you help me understand a bit more about why you've classified this as a fix?

Because you use ip-address as hosts in inventory file in etcd_cluster group. Normally, records in group writes like a short hostname or FQDN. You are right, it is not a mistake, because ETCD config require IP-address instead hostname.
My PR allow use hostnames in etcd_cluster group, and get their ip via pg_cluster_default role.

> It would be great if you could describe the details of what is offered in this PR.

I wrote some logic, which allow automatically generate variables and properties for configs. My solutions in this PR inspired by [Kubespray role](https://github.com/kubernetes-sigs/kubespray/tree/master/roles/kubespray-defaults) 
Main goal of this approach - do not write few long vars/* files. Instead, you can write few roles and drop down the monolith approach. I think, kubespray and your role are similar, and there is good opportunity to reuse their practices.
My environment VM's have few ethernet interfaces (for local subnet and management). And currently there is no way, how to use normally hostnames in etcd_cluster and set the IP var with custom interface. My PR also add this ability. 
 
> Could you share your thoughts on why this approach was chosen over defining variables in vars/main.yml?

Yeah. I think, better use simple role, which prepare final variables, then write this logic in vars/ files

vitabaks:
@dtrdnk Thanks for the answer! 

I'll review the use of a role to define variables and assess whether it brings any advantages or conveniences. I'll get back to you with my thoughts soon. At first glance, it doesn't appear to be a conventional method for defining variables.

Could you consider this change in a broader context? Perhaps you'd be interested in helping to rethink the logic for all host groups, not just for etcd. As it stands, it seems like this might be only a partial solution.

In the Consul role, we use a different approach to define the listening IP address, not using the 'inventory_hostname' variable. Here are the relevant sections for reference:
- [Consul defaults](https://github.com/vitabaks/postgresql_cluster/blob/1.10.0/roles/consul/defaults/main.yml#L124)
- [Consul config template](https://github.com/vitabaks/postgresql_cluster/blob/1.10.0/roles/consul/templates/config.json.j2#L21)

I was thinking of implementing a similar approach. What are your thoughts on this?

vitabaks:
@ThomasSanson I would be grateful if you would take a look at this PR and take part in our discussion.

dtrdnk:
> Could you consider this change in a broader context? 

Yeah. I offer two variants for using automatically generated variables:
 - globally, if the variable will be used in two roles, as is the case with ETCD.
 The address is used both in the ETCD role itself and in Patroni. For global variant using pg_cluster_defaults role.
 - locally in defaults/main.yml, if only automatically generated variables will be used in this role
        
> I was thinking of implementing a similar approach for all host groups. What are your thoughts on this?
        
With Consul, you took approximately the same approach as I did with ETCD, but did not include the connecting address in the Patroni config. You can globally define the `ip` variable, and thereby automatically determine the interface on which the service will be launched. This is exactly the approach in the kubespray:
- https://github.com/kubernetes-sigs/kubespray/blob/v2.24.1/inventory/sample/inventory.ini#L1
- https://github.com/kubernetes-sigs/kubespray/blob/v2.24.1/roles/kubernetes/node/defaults/main.yml#L3

P.S
Please, try this Vagrantfile for my work tasks
- https://github.com/dtrdnk/vagrant-powerdns-patroni/tree/3d138cfd76b1c40b681cbce86c58c14460e87537

And after successfully running Vagrant, check the   `.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory`.
You can see `ip` var in inventory, which will reused in `postgresql_cluster` role.

vitabaks:
@dtrdnk What method do you think will be more convenient?

dtrdnk:
Take both. As I say above, in roles without reusing vars between roles, just keep them in **<role_name>/defaults/main.yml**.
In roles with sharing vars, use specific default role

p.s thx for your response

vitabaks:
@dtrdnk Could you close this PR and prepare a new one that would implement this functionality for all cluster components and not just etcd?

Important! It is necessary to leave backward compatibility so that if the `ip` variable is not defined, then use the `inventory_hostname` value as it currently works.

Issue #657, opened 2024-05-13
Wal-g backup and restore command doenst work Ubuntu 22.04 due syntax issue 


garry-t:
It is not clear why it doesnt work. Problem seems like here `'%{http_code}'`. I see that it marked was tested in Ubuntu 22.04. 
Was wal-g also tested? 

https://github.com/vitabaks/postgresql_cluster/blob/f59badf3513b3b7eac2ec4edfa11ff1ef72ac4ed/vars/main.yml#L483-L488
Tried with :
/bin/bash
/bin/sh 
No reason, same issue. 
syslog output `CMD (sh -c 'curl -o /dev/null -s -w ")`

vitabaks:
Hi @garry-t 

I will be grateful for PR if there is a mistake here and you manage to solve it.

I usually use pgbackrest and a simple check, for example:

```
if [ $(psql -tAXc 'select pg_is_in_recovery()') = 'f' ]; then pgbackrest --type=full --stanza={{ pgbackrest_stanza }} backup; fi
```

vitabaks:
please attach the task file located in `/etc/cron.d/walg`

garry-t:
Ok, understood. 

```
cat /etc/cron.d/walg 
#Ansible: WAL-G: Create daily backup
30 3 * * * postgres  [ $(curl -s -o /dev/null -w '%{http_code}' http://IP:8008) = '200' ] && wal-g --config /var/lib/postgresql/.walg.json backup-push /var/lib/postgresql/15/main > /var/log/postgresql/walg_backup.log 2>&1
#Ansible: WAL-G: Delete old backups
30 6 * * * postgres  [ $(curl -s -o /dev/null -w '%{http_code}' http://IP:8008) = '200' ] && wal-g --config /var/lib/postgresql/.walg.json delete retain FULL 4 --confirm > /var/log/postgresql/walg_delete.log 2>&1
```
So I'll think how to solve it.

PR #658, opened 2024-05-13
Syntax problem with cron wal-g command

garry-t:
Fix syntax issue inside cron file #657 
Plus fix problem with wal-g run, config file now specified in command directly. 

This is a simplest solution I found. Minimum changes required, but now it works

### DIFF
--- a/molecule/tests/variables/asserts/wal_g_cron_jobs.yml
+++ b/molecule/tests/variables/asserts/wal_g_cron_jobs.yml
@@ -24,10 +24,10 @@
   run_once: true
   ansible.builtin.assert:
     that:
-      - wal_g_backup_command[0] == "[ $(curl -s -o /dev/null -w '%{http_code}' http://{{ inventory_hostname }}:{{ patroni_restapi_port }}) = '200' ]"
-      - wal_g_backup_command[1] == " && wal-g backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1"
-      - wal_g_delete_command[0] == "[ $(curl -s -o /dev/null -w '%{http_code}' http://{{ inventory_hostname }}:{{ patroni_restapi_port }}) = '200' ]"
-      - wal_g_delete_command[1] == " && wal-g delete retain FULL 4 --confirm > {{ postgresql_log_dir }}/walg_delete.log 2>&1"
+      - wal_g_backup_command[0] == "curl -I -s http://{{ inventory_hostname }}:{{ patroni_restapi_port }} | grep 200"
+      - wal_g_backup_command[1] == " && wal-g --config {{ postgresql_home_dir }}/.walg.json backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1"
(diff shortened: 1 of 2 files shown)


molecule/tests/variables/asserts/wal_g_cron_jobs.yml
```diff
@@ -24,10 +24,10 @@
   run_once: true
   ansible.builtin.assert:
     that:
-      - wal_g_backup_command[0] == "[ $(curl -s -o /dev/null -w '%{http_code}' http://{{ inventory_hostname }}:{{ patroni_restapi_port }}) = '200' ]"
-      - wal_g_backup_command[1] == " && wal-g backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1"
-      - wal_g_delete_command[0] == "[ $(curl -s -o /dev/null -w '%{http_code}' http://{{ inventory_hostname }}:{{ patroni_restapi_port }}) = '200' ]"
-      - wal_g_delete_command[1] == " && wal-g delete retain FULL 4 --confirm > {{ postgresql_log_dir }}/walg_delete.log 2>&1"
+      - wal_g_backup_command[0] == "curl -I -s http://{{ inventory_hostname }}:{{ patroni_restapi_port }} | grep 200"
+      - wal_g_backup_command[1] == " && wal-g --config {{ postgresql_home_dir }}/.walg.json backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1"
```
vitabaks, line 28: `.walg.json` is located in the postgres user's HOME and is used by default without explicitly specifying the --config option, so this is unnecessary.

garry-t:
> @garry-t why not simplify the [check](https://github.com/vitabaks/postgresql_cluster/blob/master/vars/main.yml#L586) for the same as for pgbackrest?
> 
> ```shell
> if [ $(psql -tAXc 'select pg_is_in_recovery()') = 'f' ]; then wal-g backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1; fi
> ```
> 
> or
> 
> ```shell
> [ $(psql -tAXc 'select pg_is_in_recovery()') = 'f' ] && wal-g backup-push {{ postgresql_data_dir }} > {{ postgresql_log_dir }}/walg_backup.log 2>&1
> ```
> 
> I don't see the need for different approaches.

I kept solution closer to original. Will check your suggestion.

molecule/tests/variables/asserts/wal_g_cron_jobs.yml
garry-t: Agree.

garry-t: fixed.

rrrru:
Debian bookworm
/bin/sh: 1: wal-g: not found

I think I should use the full path, like /usr/local/bin/wal-g

vitabaks:
Feel free to suggest a PR.

rrrru:
https://github.com/vitabaks/postgresql_cluster/pull/679

PR #673, opened 2024-06-06
Replacing restart pgbouncer with reload pgbouncer

SDV109:
Replacing restart pgbouncer with reload pgbouncer, in tasks for changing pgbouncer.ini, since when changing the configuration file, restarting pgbouncer is not required.


vitabaks:
Changing some parameters may still require a restart, but I agree that basically a restart is not needed. 

Thansk @SDV109

PR #679, opened 2024-06-23
Add wal_g_path variable; fix "wal-g: not found" (issue 658)

rrrru:
Debian bookworm
/bin/sh: 1: wal-g: not found


vitabaks:
Thanks!

PR #687, opened 2024-07-03
confd: haproxy: optional log-format

jimnydev:
New optional var for haproxy log-format. Supplied with an example of JSON structured log format (commented out by default).

### DIFF
--- a/roles/confd/templates/haproxy.tmpl.j2
+++ b/roles/confd/templates/haproxy.tmpl.j2
@@ -12,6 +12,10 @@ global
 defaults
     mode               tcp
     log                global
+    option             tcplog
+{% if haproxy_log_format is defined %}
+    log-format         '{{ haproxy_log_format }}'
+{% endif %}
     retries            2
     timeout queue      5s
     timeout connect    5s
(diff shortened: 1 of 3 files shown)


vitabaks:
ok, thanks @jimnydev

roles/confd/templates/haproxy.tmpl.j2
```diff
@@ -55,6 +55,9 @@ listen master_direct
     maxconn {{ haproxy_maxconn.master }}
     option tcplog
     option httpchk OPTIONS /primary
+{% if haproxy_log_format is defined %}
```
vitabaks, line 58: @jimnydev let's duplicate to the haproxy conf [file](https://github.com/vitabaks/postgresql_cluster/blob/master/roles/haproxy/templates/haproxy.cfg.j2) in case someone doesn't want to use confd service.

jimnydev: Of course, updated. Thanks!

vitabaks:
The log-format option can be specified once globally to apply to all listeners if it is used in the defaults section.

We don't need to duplicate the log-format in each listen section. This will simplify the configuration and reduce the amount of repetitive code.

@jimnydev Please check it out.

jimnydev:
Sure, I initially tried to implement this with least changes to existing setup.

Defining `log-format` within defaults works, but only if we also deduplicate `option tcplog` as well. I've reset the branch and force-pushed suggested implementation.

Thanks!

jimnydev:
Of course, pushed an update for that. I used my IDE to (un)comment out the block - and it automatically updates identation for yaml as well - I guess that's why I missed it. Thank you!

SDV109:
@jimnydev Oh, this is very strange and sad, because if users do the simplest thing, just uncomment this block, they will get an error.
@vitabaks FYI

vitabaks:
Thanks @SDV109 @jimnydev

PR #689, opened 2024-07-05
Add firewalld support for RedHat

redy01:
https://github.com/vitabaks/postgresql_cluster/discussions/501


vitabaks:
Hi @redy01 


In this case, it is necessary to support both types:

- firewalld (RHEL based)
- ufw (Debian based)

iptables by default as it is suitable for both distribution families.

deploy_pgcluster.yml
```diff
@@ -97,19 +96,19 @@
       when: ansible_os_family == "Debian"
 
     # Ansible requires the iproute package for network facts to be populated
-    - name: Make sure that the iproute is installed
+    - name: Make sure that the {{ firewall_type }} is installed
       ansible.builtin.package:
-        name: iproute
+        name: "{{ firewall_type }}"
```
vitabaks, line 101: we don't have an `iproute` value for `firewall_type` variable.

PR #713, opened 2024-08-01
Automatically configure "vm.nr_hugepages"

vitabaks:
Issue: https://github.com/vitabaks/postgresql_cluster/issues/697

Automatically configure "`vm.nr_hugepages`" for shared_buffers of 8GB or more.

If '`sysctl_set`' is '`true`', "vm.nr_hugepages" is undefined or insufficient in `sysctl_conf`, and "huge_pages" is not 'off' in `postgresql_parameters`.


rcknr:
What's the formula to calculate number of huge pages?

vitabaks:
> What's the formula to calculate number of huge pages?

the size of shared_buffers +1GB converted to kb is divided by the size of the pages in kb
```yaml
       huge_pages_required: >-
          {{
            ((shared_buffers_gb | default(0) | int + (additional_huge_pages_gb | default(1)))
            * 1024 * 1024) // huge_page_size.stdout | default(2048) | int
          }}
```

Issue #737, opened 2024-08-21
Possibility of adding privileges to objects


abyss-ms:
Hi @vitabaks 
I have a question regarding the possibility of adding privileges for users. In the postgresql_cluster project, we can find roles for adding users, databases, and schemas, but there doesn't seem to be an option to add privileges or create a role with specific privileges. Sometimes, it is necessary to create a role (or add a specific privilege to user) like pg_ls_waldir() for example for monitoring purposes. Additionally, some users or service accounts should not have full read/write access to the database.

Is it not possible to add privileges, or is there a solution that I might have missed?

Thank you in advance for your response.

vitabaks:
Hi @abyss-ms 

For this task, we can create roles to perform modules such as [postgresql_owner](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_owner_module.html#ansible-collections-community-postgresql-postgresql-owner-module), [postgresql_membership](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_membership_module.html#ansible-collections-community-postgresql-postgresql-membership-module) and [postgresql_privs](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_privs_module.html#ansible-collections-community-postgresql-postgresql-privs-module) modules. Or [postgresql_query](https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_query_module.html#ansible-collections-community-postgresql-postgresql-query-module) as a more versatile option.

in the meantime, you have the opportunity to use `post_deploy_command` [variable](https://github.com/vitabaks/postgresql_cluster/blob/master/automation/vars/system.yml#L240), in which you can specify any script / command that will be executed after the cluster deployment.

abyss-ms:
Cool. Thank you for your reply. Nice to know about post_deploy_command variable - I haden't noticed it :)
I will try to prepare something regarding postgresql_privs soon then.

PR #740, opened 2024-08-27
Possibility of adding privileges to objects

abyss-ms:
issue #737

### DIFF
--- a/automation/deploy_pgcluster.yml
+++ b/automation/deploy_pgcluster.yml
@@ -380,6 +380,9 @@
     - role: postgresql-schemas
       when: is_master | bool and postgresql_schemas | length > 0
 
+    - role: postgresql-privs
+      when: is_master | bool and postgresql_databases | length > 0
+
     - role: postgresql-extensions
       when: is_master | bool and postgresql_extensions | length > 0
 
--- a/automation/roles/postgresql-privs/tasks/main.yml
+++ b/automation/roles/postgresql-privs/tasks/main.yml
@@ -18,5 +18,6 @@
   when:
     - postgresql_privs | default('') | length > 0
     - item.role | default('') | length > 0
+    - item.db | default('') | length > 0
     - patroni_standby_cluster.host | default('') | length < 1  # do not perform on the Standby Cluster leader
   tags: postgresql_privs
--- a/automation/roles/postgresql-privs/tasks/main.yml
+++ b/automation/roles/postgresql-privs/tasks/main.yml
@@ -1,0 +1,22 @@
+---
+
+- name: Grant/revoke privileges on objects
+  community.postgresql.postgresql_privs:
+    roles: "{{ item.role }}"
+    privs: "{{ item.privs }}"
+    type: "{{ item.type }}"
+    objs: "{{ item.objs }}"
+    schema: "{{ item.schema | default(omit) }}"
+    db: "{{ item.db }}"
+    state: "{{ item.state | default('present') }}"
+    login_host: "127.0.0.1"
+    login_port: "{{ postgresql_port }}"
+    login_user: "{{ patroni_superuser_username }}"
+    login_password: "{{ patroni_superuser_password }}"
+  ignore_errors: true  # noqa ignore-errors
+  loop: "{{ postgresql_privs | flatten(1) }}"
+  when:
+    - postgresql_privs | default('') | length > 0
+    - item.role | default('') | length > 0
+    - patroni_standby_cluster.host | default('') | length < 1  # do not perform on the Standby Cluster leader
+  tags: postgresql_privs
--- a/automation/vars/main.yml
+++ b/automation/vars/main.yml
@@ -208,8 +208,8 @@ postgresql_privs: []
 #  - { role: "test", privs: "SELECT,INSERT,UPDATE", type: "table", db: "test2", objs: "test" }  # grant SELECT, INSERT, UPDATE on a table to role test
 #  - { role: "test-user", privs: "ALL", type: "database", db: "test-db", objs: "test-db" }  # grant ALL on a database to role test-user
 #  - { role: "mydb-user", privs: "SELECT", type: "table", db: "mydb", objs: "my_table", schema: "my_schema" }  # grant SELECT on a table and schema
-#  - { role: "db-user", privs: "EXECUTE", type: "function", db: "database_name", objs: "pg_ls_waldir()", schema: "pg_catalog" }  # grant execute on a pg_ls_waldir() function in schema pg_catalog to role db-user
-#  - { role: "myuser, privs: "SELECT", type: "table", db: "mydatabase", objs: "my_table2", schema: "my_schema2", state: "absent" }  # revoke SELECT on a table and schema
+#  - { role: "user", privs: "EXECUTE", type: "function", db: "db1", objs: "pg_ls_waldir()", schema: "pg_catalog" }  # grant EXECUTE on a function
+#  - { role: "user, privs: "SELECT", type: "table", db: "mydb", objs: "table2", schema: "schema2", state: "absent" }  # revoke SELECT on a table2 and schema2
 #  - { role: "test, test2", privs: "CREATE", type: "database", db: "test2", objs: "test2" }  # grant CREATE on a database test2 to role test and test2
 
 
--- a/automation/vars/main.yml
+++ b/automation/vars/main.yml
@@ -205,9 +205,13 @@ postgresql_schemas: []
 # (optional) list of privileges to be granted (if not already exists) or revoked
 # https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_privs_module.html#examples
 postgresql_privs: []
-#  - { role: "mydb-user", privs: "SELECT", type: "table", db: "mydb", objs: "my_table", schema: "my_schema" }
-#  - { role: "db-user", privs: "EXECUTE", type: "function", db: "database_name", objs: "pg_ls_waldir()", schema: "pg_catalog" }
-#  - { role: "myuser, privs: "SELECT", type: "table", db: "mydatabase", objs: "my_table2", schema: "my_schema2", state: "absent" }
+#  - { role: "test", privs: "SELECT,INSERT,UPDATE", type: "table", db: "test2", objs: "test" }  # grant SELECT, INSERT, UPDATE on a table to role test
+#  - { role: "test-user", privs: "ALL", type: "database", db: "test-db", objs: "test-db" }  # grant ALL on a database to role test-user
+#  - { role: "mydb-user", privs: "SELECT", type: "table", db: "mydb", objs: "my_table", schema: "my_schema" }  # grant SELECT on a table and schema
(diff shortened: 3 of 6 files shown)


automation/deploy_pgcluster.yml
```diff
@@ -380,6 +380,9 @@
     - role: postgresql-schemas
       when: is_master | bool and postgresql_schemas | length > 0
 
+    - role: postgresql-privs
+      when: is_master | bool and postgresql_databases | length > 0
```
vitabaks, line 384:
```suggestion
      when: is_master | bool and postgresql_privs | default('') | length > 0
```

automation/vars/main.yml
```diff
@@ -202,6 +202,12 @@ postgresql_databases: []
 postgresql_schemas: []
 #  - { schema: "myschema", db: "mydatabase", owner: "mydb-user" }
 
+# (optional) list of privileges to be granted (if not already exists) or revoked
```
vitabaks, line 205:
```suggestion
# (optional) list of privileges to be granted (if not already exists) or revoked
# https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_privs_module.html#examples
```

Let's give a link to the examples.

vitabaks:
@SDV109 Please check.

SDV109:
As for select/insert/update, there are no problems
```
test2=# \c test2 test
Test user password:
psql (16.2 (Debian 16.2-1.pgdg110+2), server 15.7 (Debian 15.7-1.pgdg110+1))
You are connected to the "test2" database as a "test" user.
test2=> select * from test;
ERROR:  permission denied for table test
test2=> select * from test;
 id |                     about                      | age
----+------------------------------------------------+-----
  1 | A cool player. A cool player. My number is 876 |  10
  2 | A cool player. A cool player. My number is 689 |  11
  3 | A cool player. A cool player. My number is 536 |  23
  4 | A cool player. A cool player. My number is 704 |  22
  5 | A cool player. A cool player. My number is 648 |  28
  6 | A cool player. A cool player. My number is 136 |  27
  7 | A cool player. A cool player. My number is 300 |  12
  8 | A cool player. A cool player. My number is 106 |  29
  9 | A cool player. A cool player. My number is 899 |  21
 10 | A cool player. A cool player. My number is 209 |  23
(10 lines)

test2=> update test set age='26' where id='4';
UPDATE 1
test2=> select * from test;
 id |                     about                      | age
----+------------------------------------------------+-----
  1 | A cool player. A cool player. My number is 876 |  10
  2 | A cool player. A cool player. My number is 689 |  11
  3 | A cool player. A cool player. My number is 536 |  23
  5 | A cool player. A cool player. My number is 648 |  28
  6 | A cool player. A cool player. My number is 136 |  27
  7 | A cool player. A cool player. My number is 300 |  12
  8 | A cool player. A cool player. My number is 106 |  29
  9 | A cool player. A cool player. My number is 899 |  21
 10 | A cool player. A cool player. My number is 209 |  23
  4 | A cool player. A cool player. My number is 704 |  26
(10 lines)
```

abyss-ms:
Hi @SDV109 

> Also, I want to note that the assignment of the create right does not work either. I must say right away that when we select the database type, we do not need to specify the schema, otherwise there will be an error

Yes, when you specify the type as "database," you cannot provide a schema. Simply omit it. It should be like this: 
`- { role: "test", privs: "CREATE", type: "database", db: "test2", objs: "test2" }`
schema is optional.

In the following case: 
`- { role: "test", privs: "CREATE", type: "database", db: "test2", objs: "", schema: "" }`
You might end up with something like this (ignoring the fact that database type cannot be used with 'schema'): `GRANT ALL PRIVILEGES ON DATABASE '' TO test`, since you haven't provided a valid database (objs).

You guys want me to add additional checks/asserts that for example when someone provide type database and provide schema then should get error or info that it is now allowed?

vitabaks:
Let's just provide more examples that may be needed most often.

SDV109:
@abyss-ms Hi, сhange the 2nd example, from grant create and al from type database to type schema. In PostgreSQL 15, when assigning rights to the database, an error of schema unavailability will be displayed. Here are the correct options that I tested:
```
- { role: "test", privs: "CREATE", type: "schema", db: "test3", objs: "public", schema: "" }
- { role: "test", privs: "ALL", type: "schema", db: "diamond", objs: "public", schema: "" }
``` 
In the example I gave, the create and ALL rights are assigned in the public scheme only for those databases to which the connection is specified, the privileges granted in other databases will not work.

SDV109:
@abyss-ms Example, I created a new gnome user and gave him create rights in the public schema of the gnome database:
```
changed: [192.168.62.192] => (item={'role': 'gnome', 'privs': 'CREATE', 'type': 'schema', 'db': 'gnome', 'objs': 'public'})
``` 
After that, I connect to the gnome database under it and can create a table, but when connecting to other databases, this does not work
```
You are connected to the "gnome" database as a "gnome" user.
gnome=> CREATE TABLE test (id INT, about TEXT, age INT);
CREATE TABLE
You are connected to the "testdb" database as a "gnome" user.
testdb=> CREATE TABLE test (id INT, about TEXT, age INT);
ERROR:  permission denied for schema public
LINE 1: CREATE TABLE test (id INT, about TEXT, age INT);
``` 
I think the right solution would be to give exactly working examples in vars/main. But if I grant the create privilege to the database, we will see an error

```
changed: [192.168.62.192] => (item={'role': 'gnome', 'privs': 'CREATE', 'type': 'database', 'db': 'testdb', 'objs': 'testdb'})
You are connected to the "testdb" database as a "gnome" user.
testdb=> CREATE TABLE test (id INT, about TEXT, age INT);
ERROR:  permission denied for schema public
LINE 1: CREATE TABLE test (id INT, about TEXT, age INT);
```

SDV109:
@abyss-ms And finally, do you have any ideas how to grant the following rights: "GRANT SELECT ON ALL TABLES IN SCHEMA public TO dba;"
According to the ansible documentation, there should be a similar scheme:
```
- name: GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA common TO caller
  community.postgresql.postgresql_privs:
    type: function
    state: present
    privs: EXECUTE
    roles: caller
    objs: ALL_IN_SCHEMA
    schema: common
``` 
Here is my vars/main, but I get the following error:
```
vars/main:
  - { role: "read", privs: "SELECT", type: "table", objs: "ALL_IN_SCHEMA", schema: "public" }
  
 log_ansible:
 failed: [192.168.62.192] (item={'role': 'read', 'privs': 'SELECT', 'type': 'table', 'objs': 'ALL_IN_SCHEMA', 'schema': 'public'}) => {"ansible_loop_var": "item", "changed": false, "item": {"objs": "ALL_IN_SCHEMA", "privs": "SELECT", "role": "read", "schema": "public", "type": "table"}, "msg": "missing required arguments: database"}
```

vitabaks:
> "missing required arguments: database"

it looks like you should always specify the database name

vitabaks:
image
image

Doc: https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_privs_module.html#ansible-collections-community-postgresql-postgresql-privs-module

automation/roles/postgresql-privs/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+- name: Grant/revoke privileges on objects
+  community.postgresql.postgresql_privs:
+    roles: "{{ item.role }}"
```
vitabaks, line 5:
```suggestion
    roles: "{{ item.role | default('postgres') }}"
```

automation/roles/postgresql-privs/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    privs: "{{ item.privs }}"
+    type: "{{ item.type }}"
+    objs: "{{ item.objs }}"
+    schema: "{{ item.schema | default(omit) }}"
+    db: "{{ item.db }}"
```
vitabaks, line 10:
```suggestion
    db: "{{ item.db | default('postgres') }}"
```

automation/roles/postgresql-privs/tasks/main.yml
abyss-ms: Hi @vitabaks that setting `"{{ item.db | default('postgres') }}"` should be ok since objs is required always in this role, that `"{{ item.role | default('postgres') }}"` could assign permissions to the wrong user, so `- item.role | default('') | length > 0` should be enough. What do you think?

automation/roles/postgresql-privs/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+    state: "{{ item.state | default('present') }}"
+    login_host: "127.0.0.1"
+    login_port: "{{ postgresql_port }}"
+    login_user: "{{ patroni_superuser_username }}"
+    login_password: "{{ patroni_superuser_password }}"
+  ignore_errors: true  # noqa ignore-errors
+  loop: "{{ postgresql_privs | flatten(1) }}"
+  when:
+    - postgresql_privs | default('') | length > 0
+    - item.role | default('') | length > 0
```
vitabaks, line 20:
```suggestion
    - item.role | default('') | length > 0
    - item.db | default('') | length > 0
```

or so, in this case, the task will not be completed at all if the necessary parameters are not specified, and not errors. Otherwise, we will need to implement the checks in the pre-checks. 

I think it is possible to describe in the comment that "db" and "role" are required.

automation/roles/postgresql-privs/tasks/main.yml
vitabaks: Ok, new [comment](https://github.com/vitabaks/postgresql_cluster/pull/740#discussion_r1738163943)

SDV109:
@vitabaks Yes, I understand, but here I cited an error in the ansible documentation, in it, as I gave an example from the off documentation, db should not be passed, although this probably only works with functions, and when it comes to tables, you need to grant select rights at the database level.

But I ran into another problem, it turns out that granting rights to select in all tables in the public schema is not enough for the user to make a select in all databases. 

So, I think it's worth describing in vars/main before the postgresql_privs: [] block that if a user wants to grant select/insert/update rights for a specific table, then this must be done at the database level and is assigned only to one database:
#  - { role: "test", privs: "SELECT,INSERT,UPDATE", type: "table", db: "test2", objs: "test" }  # grant SELECT, INSERT, UPDATE on a table to role test

And if the user wants to grant rights to the entire database or the ability to create tables, in this case it is necessary to work at the schema level indicating the schema where he wants to make changes.:
  - { role: "test", privs: "ALL", type: "schema", db: "diamond", objs: "public", schema: "" } # grant ALL PRIVILEGES  on database to role test
  - { role: "test", privs: "CREATE", type: "schema", db: "gnome", objs: "public" } # grant CREATE   on database to role test
  
  @abyss-ms FYI

abyss-ms:
Hi @SDV109 
so from your tests something like this (grant ALL on a test-db database to role test-user):
`- { role: "test-user", privs: "ALL", type: "database", db: "test-db", objs: "test-db" }`
doesn't work?

automation/roles/postgresql-privs/tasks/main.yml
abyss-ms: Done.

PR #777, opened 2024-10-02
pg_upgrade.yml: Add ‘string’ filter for postgresql version variables

ruslanloman:
# Description

Ansible pg_upgrade.yml pre_check is failed when postgresql_version is int, because
regex_replace ansible filter expects a string or pattern.

Initially postgresql_version was a string, but changed in this [PR](https://github.com/vitabaks/postgresql_cluster/commit/1cf159e97ef1f51142184b3c3bf1a39641d660ce) to int 

```
ansible-playbook pg_upgrade.yml -e "pg_old_version=15 pg_new_version=16"

TASK [upgrade : Make sure that the old and new data and config directories do not match] ***********************************************
fatal: [192.168.40.135]: FAILED! => {"msg": "The conditional check '(pg_old_datadir == pg_new_datadir) or (pg_old_confdir == pg_new_confdir)' failed. The error was: An unhandled exception occurred while templating '{{ postgresql_data_dir | regex_replace('(/$)', '') | regex_replace(postgresql_version, pg_old_version) }}'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Unexpected templating type error occurred on ({{ postgresql_data_dir | regex_replace('(/$)', '') | regex_replace(postgresql_version, pg_old_version) }}): first argument must be string or compiled pattern. first argument must be string or compiled pattern\n\nThe error appears to be in '/home/loman/postgresql_cluster/automation/roles/upgrade/tasks/pre_checks.yml': line 14, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# Stop, if the directories of the old and new versions are the same\n- name: \"Make sure that the old and new data and config directories do not match\"\n  ^ here\n"}
```

### DIFF
--- a/.github/workflows/molecule.yml
+++ b/.github/workflows/molecule.yml
@@ -66,7 +66,7 @@ jobs:
         run: make bootstrap-dev
 
       - name: Run Molecule tests
-        run: make molecule-test-scenario MOLECULE_SCENARIO="pg_upgrade"
+        run: make molecule-test
         env:
           PY_COLORS: "1"
           ANSIBLE_FORCE_COLOR: "1"
--- a/.github/workflows/molecule.yml
+++ b/.github/workflows/molecule.yml
@@ -66,7 +66,7 @@ jobs:
         run: make bootstrap-dev
 
       - name: Run Molecule tests
-        run: make molecule-test
+        run: make molecule-test-scenario MOLECULE_SCENARIO="pg_upgrade"
         env:
           PY_COLORS: "1"
           ANSIBLE_FORCE_COLOR: "1"
--- a/automation/molecule/default/converge.yml
+++ b/automation/molecule/default/converge.yml
@@ -16,7 +16,7 @@
         dcs_type: "{{ ['etcd', 'consul'] | random }}" # Set 'dcs_type' to either 'etcd' or 'consul' randomly
         consul_node_role: server # if dcs_type: "consul"
         consul_bootstrap_expect: true # if dcs_type: "consul"
-        postgresql_version: "16" # to test custom WAL dir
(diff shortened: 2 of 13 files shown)


vitabaks:
Thanks! @ruslanloman 

It is better to add a string filter to the appropriate parts of the code than to require users to specify the version as a string. This will be a more flexible and more reliable option. 

example

```yaml
{{ postgresql_data_dir | string | regex_replace('(/$)', '') | regex_replace(postgresql_version | string, pg_old_version | string) }}
```

vitabaks:
@ruslanloman Are you planning a new PR or can I do it myself?

ruslanloman:
Hi @vitabaks,

Thank you for your comment. Yes, I'm preparing a new PR. I agree about this part "require users to specify the version as a string". I found out that we can use jinja [replace](https://tedboy.github.io/jinja2/templ14.html#replace) filter instead of replace_regex. I updated code, but checking that deployment and upgrade works as expected.

automation/vars/main.yml
```diff
@@ -173,7 +173,7 @@ consul_services:
 
 
 # PostgreSQL variables
-postgresql_version: 16
+postgresql_version: 15
```
vitabaks, line 176:
```suggestion
postgresql_version: 16
```

ruslanloman:
@vitabaks Updated. My local deployment/upgrade test passed.

ruslanloman:
> recommended to use a string filter to avoid possible errors, example:
> 
> ```
> replace(postgresql_version | string, pg_old_version | string)
> ```

Updated. Thanks

PR #808, opened 2024-11-14
Fix maximum_lag_on_failover case when value set to 0

garry-t:
- fix case when set 0 for maximum_lag_on_failover


vitabaks:
Could you attach an example of the error that we are correcting? 

I don't think anyone will set 0 as this will turn off autofailover at the slightest replication lag

vitabaks:
In this case, just enable synchronous replication.

PR #828, opened 2024-12-06
patroni_replication_username in hba

allig4t0r:
Across many versions there is such line in postgresql_pg_hba:
`  - { type: "local", database: "replication", user: "{{ patroni_replication_username }}", address: "", method: "trust" }`

I assume it's a typo and it needs to be:
`  - { type: "local", database: "replication", user: "{{ patroni_replication_username }}", address: "", method: "trust" }`

This is the only critical place as far as I can see but I'd highly suggest to double-check me.


vitabaks:
Thanks @allig4t0r

PR #841, opened 2024-12-13
add ability to set some variables without manually edit vars file

zappbrannigan34:
just add jinja template with default values instead hardcoded values for some variables. for use extra variables with ansible-playbook run or environment variables


vitabaks:
Thank you for your contribution! However, I’m not sure if this change is necessary. The variables in question are not hardcoded — they can already be overridden by modifying their values in the variables file or by using --extra-vars, which is exactly how we manage configurations in Autobase.

Could you elaborate on the specific scenario where this approach would not suffice? It would help me better understand the need for this adjustment.

PR #853, opened 2024-12-19
Auto set the number of max_worker_processes and max_parallel_workers

klention:
This PR introduces dynamic configuration for `max_worker_processes` and `max_parallel_workers` based on the number of vCPUs on the server:
- max_worker_processes: set to the higher value between the number of vCPUs or the minimum value of 16.
- max_parallel_workers: set to the higher value between half the number of vCPUs or the minimum value of 8.

This ensures optimal configuration for both small and large servers, adapting to the available resources while maintaining reasonable defaults.


vitabaks:
Thanks @klention

Issue #822, opened 2024-11-29
Authentication support in etcd


chobostar:
I would like to discuss the current lack of [certificate authentication](https://etcd.io/docs/v3.2/op-guide/security/) in etcd configuration:

https://github.com/vitabaks/autobase/blob/master/automation/roles/etcd/templates/etcd.conf.j2

In a previous discussion, it was mentioned that sensitive information is not stored in DCS, as highlighted here:
https://github.com/vitabaks/autobase/issues/361#issuecomment-1572279916
>  sensitive information is not stored in DCS.

Despite this position, there remains a potential risk scenario:
* A malicious `archive_command` can be set in DCS. This could potentially lead to Remote Code Execution (RCE) exploits, allowing unauthorized access to sensitive data, e.g. an attacker could create a superuser without a password, modify `pg_hba.conf` to `trust`, or corrupt pgdata.
* It is theoretically possible to remove data, bootstrap a new cluster, and direct existing nodes to connect to a malicious master using valid replication passwords.

Given the potential risks outlined, I kindly request reconsideration of the current decision to leave etcd unprotected. Implementing certificate authentication can provide an additional layer of security and protect against these vulnerabilities. 

Thanks for attention and waiting for your feedback for possible security mitigation.

vitabaks:
Thank you so much for raising this important issue! We fully agree that security is a critical aspect, and we’ll prioritize this task accordingly.

As an open-source project, we greatly appreciate contributions from the community. If you have the time and interest, please feel free to submit a PR with your proposed solution. This will help us implement the necessary changes faster and improve the project even further.

vitabaks:
@klention you might be interested in the implementation of this functionality: certificate authentication in etcd.

klention:
> @klention you might be interested in the implementation of this functionality: certificate authentication in etcd.

Will have a look at it.

vitabaks:
Done https://github.com/vitabaks/autobase/pull/857

PR #849, opened 2024-12-15
Enabled TLS encryption for PostgreSQL and PgBouncer

klention:
Enabled TLS encryption for PostgreSQL and PGBouncer (related issue: https://github.com/vitabaks/autobase/issues/361)

Generate a Self Signed OpenSSL certificate.

PR #857, opened 2024-12-21
Enabled TLS encrypted communication for etcd

klention:
Issue: https://github.com/vitabaks/autobase/issues/822

Related PR: #849

TLS Communication Encryption is now enabled for:

- [x] Between PostgreSQL nodes 
- [x] Between PgBouncer and PostgreSQL nodes 
- [x] Between etcd nodes
- [x] Between Patroni and etcd nodes
- [x]  Between vip-manager and etcd nodes
- [x]  Between confd and etcd nodes


klention:
@vitabaks I am having some trouble catching why tests are failing on Debian 12 and Ubuntu 23.04.
So far I have tested deploying it on Rocky Linux 8, 9 and Debian 11. For those it works fine. Can you have a look for the two failed tests?

klention:
Resolved #822

klention:
#859

automation/roles/tls_certificate/generate/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+        subject_alt_name: "{{ subject_alt_name }}"
+      register: csr
+
+    - name: "Sign server certificate with the CA"
+      community.crypto.x509_certificate_pipe:
+        csr_content: "{{ csr.csr }}"
+        provider: ownca
+        ownca_path: "/etc/tls/ca.crt"
+        ownca_privatekey_path: "/etc/tls/ca.key"
+        ownca_not_after: +3650d
```
vitabaks, line 104:
```suggestion
        ownca_not_after: +{{ tls_cert_valid_days | default(3650) }}d
```

automation/roles/etcd/tasks/main.yml
```diff
@@ -86,6 +86,14 @@
     state: directory
   tags: etcd, etcd_conf
 
+- name: Fetch etcd TLS certificate, key and CA from the master node
+  ansible.builtin.include_role:
+    name: ../roles/tls_certificate/copy
+  vars:
+    copy_for: "etcd"
+  when: tls_cert_generate|bool
+  tags: etcd, etcd_conf
```
vitabaks, line 95: I think we could call this role at the `etcd_cluster.yml` playbook level rather than from the tasks file.
```
   - role: tls_certificate/copy
      vars:
        copy_for: "etcd"
```

automation/roles/pgbouncer/tasks/main.yml
```diff
@@ -36,6 +36,14 @@
     mode: "0750"
   tags: pgbouncer_conf, pgbouncer
 
+- name: Fetch PostgreSQL TLS certificate, key and CA from the master node
+  ansible.builtin.include_role:
+    name: ../roles/tls_certificate/copy
+  vars:
+    copy_for: "pg"
+  when: tls_cert_generate|bool
+  tags: pgbouncer_conf, pgbouncer
```
vitabaks, line 45: I think we could call this role at the `deploy_pgcluster.yml` playbook level rather than from the tasks file.

```
   - role: tls_certificate/copy
      vars:
        copy_for: "pg"
```

automation/roles/vip-manager/templates/vip-manager.yml.j2
```diff
@@ -64,16 +64,16 @@ etcd-user: {{ patroni_etcd_username | default("") }}
...
-{% if patroni_etcd_cert_file | default("") | length > 0 %}
+{% if tls_cert_generate | bool %}
 # when etcd-cert-file and etcd-key-file are specified, we will authenticate at the etcd endpoints using this certificate and key.
-etcd-cert-file: {{ patroni_etcd_cert_file | default("") }}
+etcd-cert-file: {{ tls_cert_path | default('/etc/tls/server.crt') }}
 {% endif %}
-{% if patroni_etcd_key_file | default("") | length > 0 %}
-etcd-key-file: {{ patroni_etcd_key_file | default("") }}
+{% if tls_cert_generate | bool %}
+etcd-key-file: {{ tls_privatekey_path | default('/etc/tls/server.key') }}
```
vitabaks, line 76: I think one condition `{% if tls_cert_generate | bool %}` is enough for all TLS-related options.

automation/deploy_pgcluster.yml
```diff
@@ -83,6 +83,15 @@
...
       tags: always
 
+    - name: Make sure that the python3-cryptography package is present
+      ansible.builtin.package:
+        name: python3-cryptography
+        state: present
+      register: pack_status
+      until: pack_status is success
+      delay: 5
+      retries: 3
```
vitabaks, line 93: Perhaps this should be part of the `tls_secrificate/generate` role.

automation/vars/main.yml
```diff
@@ -241,9 +245,11 @@ postgresql_parameters:
   - { option: "max_connections", value: "1000" }
   - { option: "superuser_reserved_connections", value: "5" }
   - { option: "password_encryption", value: "{{ postgresql_password_encryption_algorithm }}" }
-  - { option: "ssl", value: "on"}
-  - { option: "ssl_cert_file", value: "{{ tls_cert_path }}"}
-  - { option: "ssl_key_file", value: "{{ tls_privatekey_path }}"}
+  - { option: "ssl", value: "{% if tls_cert_generate | bool %}on{% else %}off{% endif %}"}
+  - { option: "ssl_prefer_server_ciphers", value: "{% if tls_cert_generate | bool %}on{% else %}off{% endif %}"}
```
vitabaks, line 249:
```suggestion
  - { option: "ssl", value: "{{ 'on' if tls_cert_generate | bool else 'off' }}" }
  - { option: "ssl_prefer_server_ciphers", value: "{{ 'on' if tls_cert_generate | bool else 'off' }}" }
```

automation/roles/etcd/tasks/main.yml
klention: I placed it there so the certificate is fetched immediately after the `/etc/etcd` directory is created. Adding it before the execution of the etcd role won't work because the `/etc/etcd` directory doesn't exist yet. On the other hand, running it after the etcd role will cause the etcd daemon to fail to start due to missing certificates.

klention: I placed it there so the certificate is fetched immediately after the `/etc/etcd` directory is created and etcd is installed (but not yet configured). Adding it before the execution of the etcd role likely won't work because etcd won't be installed yet, meaning the etcd user and group will be missing, and changing the file ownership will fail. On the other hand, running it after the etcd role will cause the etcd daemon to fail to start due to missing certificates.

automation/deploy_pgcluster.yml
klention: Agree.

automation/roles/pgbouncer/tasks/main.yml
klention: I will have to check if the same thing might apply as with etcd suggestion.

automation/roles/etcd/tasks/main.yml
vitabaks: Ok, thanks

automation/roles/pgbouncer/tasks/main.yml
klention: The will now be called from `deploy_pgcluster.yml`

automation/roles/tls_certificate/generate/tasks/main.yml
```diff
@@ -0,0 +1,122 @@
...
+    state: absent
+  loop:
+    - "{{ tls_privatekey_path | default('/etc/tls/server.key') }}"
+    - "{{ tls_cert_path | default('/etc/tls/server.crt') }}"
+    - "{{ tls_ca_cert_path | default('/etc/tls/ca.crt') }}"
+    - "{{ tls_ca_privatekey_path | default('/etc/tls/ca.key') }}"
+    - "{{ tls_etcd_cert_path | default('/etc/etcd/server.crt') }}"
+    - "{{ tls_etcd_ca_cert_path | default('/etc/etcd/ca.crt') }}"
+    - "{{ tls_etcd_privatekey_path | default('/etc/etcd/server.key') }}"
+    - "/etc/tls"
```
vitabaks, line 23:
```suggestion
```
Let’s avoid deleting the entire directory. To minimize the risk of accidentally deleting certificates unrelated to the PostgreSQL cluster, it makes sense to handle this more cautiously when converting an existing installation into a cluster. In such cases, we cannot assume in advance whether there are other certificates in the directory besides the ones we generate.

vitabaks: Or perhaps we could remove this task entirely. Could you clarify the purpose of deleting the existing certificates, if any? It would help us understand if this step is necessary or if there’s an alternative approach.

klention: The `/etc/tls` directory does not exist by default on any Debian- or RHEL-based systems. Therefore, the only files present there will be the ones we explicitly add. In this case, it’s just the certificate for the PostgreSQL cluster. This means it should safe to delete if needed.

The reason I included that task is to ensure flexibility for re-initializing the cluster. For example, if you want to reinitialize it with different hostnames or IPs, the etcd cluster will not function correctly. This is because the IPs and hostnames must be part of the certificate's Subject Alternative Name (SAN). Without this, encrypted communication between etcd nodes would fail. It will complain with messages like `Bad Certificate`

Additionally, if you plan to add new nodes to an existing cluster in the future (thereby expanding the etcd cluster), those certificates will also need to be regenerated.

vitabaks: However, since we are only deleting specific files, we should not delete the entire ‘tls’ directory.

vitabaks:
Great job @klention! Thanks a lot for the contribution.

The remaining task is to implement TLS between the [confd](https://github.com/klention/postgresql_cluster/blob/tls_etcd_communication/automation/roles/confd/templates/confd.toml.j2) service and etcd nodes to ensure full TLS coverage across all cluster components.

Configuration options:

- `client_cakeys` (string) - The client CA key file.
- `client_cert` (string) - The client cert file.
- `client_key` (string) - The client key file.

Doc: https://github.com/kelseyhightower/confd/blob/master/docs/configuration-guide.md

Let me know if you're ready to do it, or I'll set aside time for it.

klention:
> Great job @klention! Thanks a lot for the contribution.
> 
> The remaining task is to implement TLS between the [confd](https://github.com/klention/postgresql_cluster/blob/tls_etcd_communication/automation/roles/confd/templates/confd.toml.j2) service and etcd nodes to ensure full TLS coverage across all cluster components (not considering the consul cluster).
> 
> Configuration options:
> 
> * `client_cakeys` (string) - The client CA key file.
> * `client_cert` (string) - The client cert file.
> * `client_key` (string) - The client key file.
> 
> Doc: https://github.com/kelseyhightower/confd/blob/master/docs/configuration-guide.md
> 
> UPD: I'll do it.

Thank you for all of this 👍

PR #867, opened 2025-01-10
Add support for configuring Patroni REST API request queue size

chobostar:
This introduces an optional `request_queue_size` parameter for the Patroni REST API, with a default value of 5. It controls the TCP socket queue size, and once full, additional requests will receive a "Connection denied" error.

### Motivation

The proposed change addresses an availability issue in Patroni's REST API caused by the low default TCP accept queue size (`5`), which can lead to frequent listen queue overflows under high load, such as **HAProxy health checks**. By introducing a configurable `request_queue_size` option, this fix ensures greater scalability and stability for high-traffic environments while maintaining backward compatibility with the default settings

### References

https://github.com/patroni/patroni/pull/2643

https://patroni.readthedocs.io/en/latest/yaml_configuration.html#rest-api
> request_queue_size: (optional): Sets request queue size for TCP socket used by Patroni REST API. Once the queue is full, further requests get a “Connection denied” error. The default value is 5.

### DIFF
--- a/automation/roles/patroni/templates/patroni.yml.j2
+++ b/automation/roles/patroni/templates/patroni.yml.j2
@@ -34,6 +34,9 @@ restapi:
 #    username: username
 #    password: password
 {% endif %}
+{% if patroni_restapi_request_queue_size is defined %}
+  request_queue_size: {{ patroni_restapi_request_queue_size |int }}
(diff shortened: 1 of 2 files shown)


vitabaks:
Ok, thanks!

PR #945, opened 2025-04-20
Single Hetzner LB with multiple ports

PhilWaldmann:
Hello,

I've created a PR that provides a single Hetzner LB with multiple ports – instead of one LB for primary, replica, and other servers.
I'm still missing additions for AWS, Azure, and other servers. I'd like to know if the style/variable names are suitable for you.

Thanks,
Philipp

### DIFF
--- a/automation/roles/cloud_resources/defaults/main.yml
+++ b/automation/roles/cloud_resources/defaults/main.yml
@@ -32,6 +32,9 @@ database_public_allowed_ips: "" # (comma-separated list of IP addresses in CIDR 
 
 # Load balancer
 cloud_load_balancer: true # Create a Load Balancer in the Cloud.
+cloud_load_balancer_unified: false # Creates a single load balancer with different ports
+cloud_load_balancer_replica_port: 5433
+cloud_load_balancer_sync_port: 5434
 
 # Backups (if 'pgbackrest_install' or 'wal_g_install' is 'true')
 aws_s3_bucket_create: true # if 'cloud_provider=aws'
--- a/automation/roles/cloud_resources/tasks/hetzner.yml
+++ b/automation/roles/cloud_resources/tasks/hetzner.yml
@@ -533,12 +533,14 @@
         - "primary"
         - "replica"
         - "sync"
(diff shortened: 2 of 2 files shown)


vitabaks:
Hi @PhilWaldmann! Thank you for your contribution.

> I'm still missing additions for AWS, Azure, and other

It would be nice to work this out for other clouds in the future. Would you like to try to implement this?

automation/roles/cloud_resources/defaults/main.yml
```diff
@@ -32,6 +32,9 @@ database_public_allowed_ips: "" # (comma-separated list of IP addresses in CIDR
 
 # Load balancer
 cloud_load_balancer: true # Create a Load Balancer in the Cloud.
+cloud_load_balancer_unified: false # Creates a single load balancer with different ports
+cloud_load_balancer_replica_port: 5433
+cloud_load_balancer_sync_port: 5434
```
vitabaks, line 37: Maybe we should use the same ports as we do when using [haproxy](https://github.com/vitabaks/autobase/blob/master/automation/roles/common/defaults/main.yml#L25)?

```yaml
  master: 5000
  replicas: 5001
  replicas_sync: 5002
  replicas_async: 5003
```

PhilWaldmann:
I'll give it a try in the next few days!

PhilWaldmann:
Hi, unfortunately it took a while. 
I'm not sure if it's doable with AWS, Digitalocean and GCP - the health checks can't be configured per "service", so it would lose all its usefulness. I've just read the ansible docs for the relevant services.
Do you know if for example the aws classic LB is able to configure individual health checks per port?

vitabaks:
I'll take a look at it a bit later.

vitabaks:
Is this PR still relevant?

I think it will not make sense in Autobase 3.0, where it is planned to redesign the proxy layer, which will have only one port for both reading and writing.

PhilWaldmann:
If you are redesigning the proxy part anyways, it's probably not worth the effort to finish this PR.

PR #949, opened 2025-04-25
Use a more flexible Ansible dependency notation

pat-s:
Currently Ansible deps are hardcoded to a specific version.
This raises issues when installing the collection into an inventory with other roles/collections which make use of the same dependencies or if these dependencies are specified standalone.

Having dedicated versions means that all other assets must follow the ones listed in `autobase` or none will work.

I suggest to make the dependency notation more flexible by using `>=` instead of `==`. This way, users can install newer versions of the deps and in case of issues, the last tested versions are listed descriptively in the collection description.

### DIFF
--- a/automation/requirements.yml
+++ b/automation/requirements.yml
@@ -1,22 +1,22 @@
 ---
 collections:
   - name: amazon.aws
-    version: "==9.3.0"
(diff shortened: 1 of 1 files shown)


pat-s:
#950

PR #972, opened 2025-05-06
fix: move `yedit.py` to collection modules dir

pat-s:
When calling a playbook from the collection, the `yedit` tasks in the `patroni` can't be executed as the module lookup logic doesn't work with modules nested inside the roles.

To solve this, `yedit.py` must be moved to the collection module lookup directory.

This can be reproduced with 

```
ansible-playbook <collection dir>/deploy_pgcluster.yml
```

using the new collection structure in current HEAD.


vitabaks:
@pat-s 

```
  ERROR! couldn't resolve module/action 'yedit'. This often indicates a misspelling, missing collection, or incorrect module path.
  
  The error appears to be in '/home/runner/work/autobase/autobase/automation/roles/patroni/tasks/main.yml': line 816, column 7, but may
  be elsewhere in the file depending on the exact syntax problem.
  
  The offending line appears to be:
  
  
      - name: Edit patroni.dynamic.json | disable archive_command (if enabled)
        ^ here
```

vitabaks:
I had the idea to remove this plugin and instead use standard modules such as `replace` or `lineinfile`. Here, for example, we use them: https://github.com/vitabaks/autobase/blob/master/automation/roles/upgrade/tasks/update_config.yml

pat-s:
> I had the idea to remove this module and instead use standard modules such as replace or lineinfile. Here, for example, we use them: [master/automation/roles/upgrade/tasks/update_config.yml](https://github.com/vitabaks/autobase/blob/master/automation/roles/upgrade/tasks/update_config.yml?rgh-link-date=2025-05-06T16%3A51%3A47Z)

Yes this is likely the best option. I also thought initially about using the `slurp` module, but I didn't dig deep enough to see if it can catch everything.

There's also https://github.com/mitre/ansible-role-yedit as a possible upstream option which can be imported.

Let's close this here and first discuss the reorganization topic we discussed about lately.

PR #979, opened 2025-05-07
feat: add generic s3 backup options

pat-s:
So far this was only able when using `cloud-provider`. Now, one can set `backup_provider` to one of the mentioned options and backup to S3 through `pgbackrest` when the `autobase` cluster has been deployed via Ansible.

Feel free to tweak to your liking, I just had this already in use since some time and wanted to part the idea upstream.

In addition to setting `backup_provider`, this also requires setting the respective access keys, storage name and other vars required to configure the S3 backend.

### DIFF
--- a/automation/roles/common/defaults/main.yml
+++ b/automation/roles/common/defaults/main.yml
@@ -499,6 +499,9 @@ patroni_create_replica_methods:
   #  - pg_probackup
   - basebackup
 
+# if set, will configure s3 backups to the respective provider. Possible values: aws_s3, azure_s3, digitalocean_s3, hetzner_s3, gcp_s3
+backup_provider: ""
+
 pgbackrest:
   - { option: "command", value: "{{ pgbackrest_patroni_cluster_restore_command }}" }
   - { option: "keep_data", value: "True" }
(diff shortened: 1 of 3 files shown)


pat-s:
relates to #916 and is a subset of #919

vitabaks:
In fact, we’re enabling auto_conf here even when the cluster is not fully deployed in the cloud via autobase. I’m not entirely sure why this is necessary without functionality like automatic S3 bucket creation — that was the original idea behind using auto_conf.

If the user creates the bucket manually, they can also define the configuration themselves based on the example in the [documentation](https://autobase.tech/docs/management/backup#command-line). In this case, it actually makes more sense to do it manually, since the bucket name may differ from the [default](https://github.com/vitabaks/autobase/blob/2.2.0/automation/roles/cloud-resources/defaults/main.yml#L36) suggested in auto_conf tasks

I thought it would be better to finish #919, where we make it possible to deploy bucket in the right cloud (regardless of the cluster deployment method) and automatically configure the backup configuration there..

pat-s:
There are many ways to approach this. 
This was just one which I used to quickly get it going on my end. Feel free to approach it differently, just wanted to share this one, so others could possibly adapt it.

pat-s:
@vitabaks Shall we close here? From what I understood, you don't want this approach to be added, right?

Issue #751, opened 2024-09-09
Add the ability to deploy Postgres in containers


ruslanloman:
Hello,

Thank you for your great project.

Have you considered the option to deploy Postgres/Patroni/etcd/Haproxy in containers instead of installing packages.
This approach has many advantages.

1. Easier cluster upgrades without worrying about package dependencies.
2. Provides an OS-agnostic setup.
3. Use docker upstream images for postgres, etcd.

Thank you!

vitabaks:
Hello @ruslanloman 

> Have you considered the option to deploy Postgres/Patroni/etcd/Haproxy in containers instead of installing packages.

Adding Docker support to the `postgresql_cluster` project would require significant changes to the automation code. 

We believe Docker is not an ideal solution for production environments, as it introduces unnecessary abstraction and an additional maintenance point. There are already Kubernetes operators for containerized solutions, but the strength of `postgresql_cluster` lies in its independence from Docker or Kubernetes. 

This approach sets us apart from other HA solutions and, with its high-quality automation, it can serve as a solid alternative to both Kubernetes-based systems and managed cloud databases.

ruslanloman:
Hello @vitabaks

Thank you for your reply.

>  Adding Docker support to the postgresql_cluster project would require significant changes to the automation code.

The postgresql_cluster ansible code is well written, so it shouldn't be a big problem to add an option to run the process in a container.  I can help with adding such functionality as I'm very interested in it. I like the approach used in the Kubespray project, for example, for [etcd](https://github.com/kubernetes-sigs/kubespray/blob/master/roles/etcd/templates/etcd-docker.service.j2)

> We believe Docker is not an ideal solution for databases in production environments, as it introduces unnecessary abstraction and an additional maintenance point. There are already Kubernetes operators for containerized solutions, but the strength of postgresql_cluster lies in its independence from Docker or Kubernetes.

I understand your point of view and partially agree. What I would like to achieve from using Docker/containerd is an operating system-agnostic PostgreSQL installation, as well as atomic upgrades and rollbacks, without worrying about potential package dependency issues. Using Docker/containerd would be an option, like having deb/rpm/docker.

> This approach sets us apart from other HA solutions and, with its high-quality automation, it can serve as a solid alternative to both Kubernetes-based systems and managed databases from cloud providers.

[...]

PR #1035, opened 2025-05-31
WIP: Add support for deploying in containers

vitabaks:
Issue: https://github.com/vitabaks/autobase/issues/751

This PR introduces support for deploying cluster components inside Docker containers via `installation_method: docker` variable.

> [!NOTE]
> We do not recommend running production databases inside Docker containers. However, since some of our users prefer containers, we are adding this functionality as an option — but it will not be the default mode.

Our goal is to make this as flexible and transparent as possible, closely resembling traditional on-host package installations. For example:
- systemd wrappers enable familiar `systemctl start|stop patroni` commands
- CLI wrappers (e.g. `patronictl`) will internally map to `docker exec -it <container> patronictl @`

We’re still exploring implementation details, so feedback and ideas are very welcome!


vitabaks:
If you're interested in this feature, please consider becoming a [sponsor](https://autobase.tech/docs/sponsor).

The development of this feature requires sponsorship to fund developer efforts. If you're already a sponsor and are interested in this feature, please leave a comment here so we can prioritize this issue accordingly.

Issue #1083, opened 2025-06-24
[Bug] Failed checks while deploying standby cluster


lcoon21:
### Bug description

I'm deploying standby cluster which replicates from external postgresql node

I have in my all.yml

```
patroni_standby_cluster:
  host: "172.16.8.14" # an address of remote master
  port: "5432" # a port of remote master
  primary_slot_name: "standby_slot"  # which slot on the remote master to use for replication (optional)
  restore_command: "pgbackrest --stanza=example archive-get %f %p"  # command to restore WAL records from the remote master to standby leader (optional)
  create_replica_methods:
    - pgbackrest
```

And at first I tried to set the hosts like this at hosts.ini

```
[master]
172.16.8.7 hostname=db-1 postgresql_exists=false

[replica]
172.16.8.8 hostname=db-2 postgresql_exists=false 
172.16.8.9 hostname=db-3 postgresql_exists=false
```

### Expected behavior

And now I expect that tasks are going to consider this option , while my configuration includes `patroni_standby_cluster` option, but in tasks like this:

```
- block: # PITR (custom bootstrap) - password reset for PostgreSQL users
    - name: Make sure the Master is not in recovery mode
      ansible.builtin.command: >-
        {{ postgresql_bin_dir }}/psql -p {{ postgresql_port }} -U {{ patroni_superuser_username }} -d postgres -tAXc
        "select pg_is_in_recovery()"
      register: pg_is_in_recovery
      until: pg_is_in_recovery.stdout == "f"
      retries: "{{ (cluster_restore_timeout | default(86400)) | int // 30 }}" # timeout 24 hours
      delay: 30
      changed_when: false
      when: inventory_hostname == groups['master'][0]
```

It's impossible cuz when it's standby cluster then all the nodes are standby, so the condition can't be performed.

If I set in hosts.ini all the nodes as replica like this:

```
[replica]
172.16.8.7 hostname=db-1 postgresql_exists=false
172.16.8.8 hostname=db-2 postgresql_exists=false 
172.16.8.9 hostname=db-3 postgresql_exists=false
```

I see new failures because there's no object in [master] dict:

[...]

PR #1084, opened 2025-06-24
Fix: Do not perform on the Standby Cluster

vitabaks:
Issue: https://github.com/vitabaks/autobase/issues/1083

Skip tasks that require write access—such as user creation or password resets—when deploying the Standby Cluster, because these clusters are read-only and such operations will fail.


vitabaks:
@lcoon21 please test.

lcoon21:
ty 👌🏻

PR #1088, opened 2025-06-27
Update pg_ctl to pg_ctlcluster for PostgreSQL 16 and high

SDV109:
This change is necessary for the pg_upgrade playbook to work when upgrading to version 16 or higher.

When updating to version 16 or higher, we get the following error: 
```
pg_ctl: could not start server\nExamine the log output.", "stderr_lines": ["pg_ctl: could not start server", "Examine the log output."], "stdout": "waiting for server to start.... stopped waiting", "stdout_lines": ["waiting for server to start.... stopped waiting"]
```

### DIFF
--- a/automation/roles/upgrade/tasks/schema_compatibility.yml
+++ b/automation/roles/upgrade/tasks/schema_compatibility.yml
@@ -28,7 +28,7 @@
 
 - name: "Start new PostgreSQL on port {{ schema_compatibility_check_port }} to check the schema compatibility"
   ansible.builtin.command: >
-    {{ pg_new_bindir }}/pg_ctl -D {{ pg_new_datadir }}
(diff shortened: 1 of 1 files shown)


vitabaks:
Your change will break compatibility with RHEL distributions. 

I doubt the need to use pg_ctlcluster, pg_ctl should also work correctly. Try to run the pg_ctl command manually and examine the error log.

SDV109:
The PR is cancelled.

The problem was not in pg_ctl, but in the parameters passed, to be more precise, in shared_preload_libraries. The sequence of extensions is important, namely: pg_stat_statements,pg_stat_cache. 
If the order is reversed, pg_ctl will not run PostgreSQL version 16 and higher.

Issue #1127, opened 2025-07-11
[Bug] consul_cluster.yml playbook fails when check installed packages on localhost


evgeny-bushuev-incountry:
### Bug description

```
TASK [Make sure the python3-pip package are present on the control host: consul_cluster.yml] **********************
Friday 11 July 2025  18:48:09 +0700 (0:00:07.051)       0:00:34.531 *********** 
FAILED - RETRYING: [test-db-10-9-12-155 -> localhost]: Make sure the python3-pip package are present on the control host (3 retries left).
FAILED - RETRYING: [test-db-10-9-12-155 -> localhost]: Make sure the python3-pip package are present on the control host (2 retries left).
FAILED - RETRYING: [test-db-10-9-12-155 -> localhost]: Make sure the python3-pip package are present on the control host (1 retries left).
task path: /Users/user/.ansible/collections/ansible_collections/vitabaks/autobase/playbooks/consul_cluster.yml:55
fatal: [test-db-10-9-12-155 -> localhost]: FAILED! => {"attempts": 3, "changed": false, "msg": "Error: No available formula with the name \"python3-pip\". Did you mean python-yq?"}
```

### Expected behavior

Works as expected

### Steps to reproduce

1.
2.
3.
...

### Installation method

Command line

### System info

Ubuntu 24.04
Ansible 11

### Additional info

Looks like there is no `ansible_distribution` fact gathered for locahost. This task is supposed to be skipped on MacOSX

vitabaks:
Feel free to create PR.

Issue #1126, opened 2025-07-11
[Bug] consul_cluster.yml fails on pre-tasks


evgeny-bushuev-incountry:
### Bug description

When execute playbook `consul_cluster.yml` there is an error:

```
TASK [Check if the consul role requirements (ansible.utils) are installed: consul_cluster.yml] ********************
Friday 11 July 2025  18:12:03 +0700 (0:00:00.058)       0:00:35.182 *********** 
task path: /Users/user/.ansible/collections/ansible_collections/vitabaks/autobase/playbooks/consul_cluster.yml:22
fatal: [test-db-10-9-12-155 -> localhost]: FAILED! => {"msg": "error occurred on host user.local: sudo password is incorrect"}
```

```
TASK [Make sure the unzip package are present on the control host: consul_cluster.yml] ****************************
Friday 11 July 2025  18:38:09 +0700 (0:00:00.016)       0:00:27.675 *********** 
task path: /Users/user/.ansible/collections/ansible_collections/vitabaks/autobase/playbooks/consul_cluster.yml:42
fatal: [test-db-10-9-12-155 -> localhost]: FAILED! => {"msg": "error occurred on host user.local: sudo password is incorrect"}
```

```
TASK [Make sure the python3-pip package are present on the control host: consul_cluster.yml] **********************
Friday 11 July 2025  18:41:58 +0700 (0:00:12.953)       0:00:39.742 *********** 
task path: /Users/user/.ansible/collections/ansible_collections/vitabaks/autobase/playbooks/consul_cluster.yml:55
fatal: [test-db-10-9-12-155 -> localhost]: FAILED! => {"msg": "error occurred on host user.local: sudo password is incorrect"}
```

### Expected behavior

Works as expected

### Steps to reproduce

1. 
2.
3.
...

### Installation method

Command line

### System info

Ubuntu 24.04
Ansible 11

### Additional info

Looks like `delegate_to: localhost` fails when `become: true`

vitabaks:
Feel free to create PR.

PR #1129, opened 2025-07-11
fix consul playbook

evgeny-bushuev-incountry:
fixes:
- #1126
- #1127

### DIFF
--- a/automation/playbooks/consul_cluster.yml
+++ b/automation/playbooks/consul_cluster.yml
@@ -8,6 +8,17 @@
   environment: "{{ proxy_env | default({}) }}"
 
   pre_tasks:
+
+    - name: Gather ansible_distribution from localhost
+      setup:
+        gather_subset:
+        - '!all'
+        - '!min'
+        - distribution
+      become: false
+      delegate_to: localhost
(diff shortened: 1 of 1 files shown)


vitabaks:
@evgeny-bushuev-incountry  Thank you so much for your contribution.

We do not use **Consul** in production, and it is not as well tested as the scheme with **etcd**,
so your corrections are useful for the project.

automation/playbooks/consul_cluster.yml
```diff
@@ -47,8 +59,11 @@
       delay: 5
       retries: 3
       delegate_to: localhost
+      become: false
       run_once: true # noqa run-once
-      when: ansible_distribution != "MacOSX"
+      when:
+        - hostvars['localhost']['ansible_distribution'] != "MacOSX"
+        - hostvars['localhost']['ansible_distribution'] != "Ubuntu"
```
vitabaks: I'm not sure if we should exclude ubuntu.

vitabaks: instead, I added `ignore_errors`.

PR #1132, opened 2025-07-12
fix: `config_pgcluster` check_mode 

pat-s:
`config_pgcluster` can't currently be executed in check mode as some tasks error out when in such and `check_mode: false` must be used to not fail at these.

Tested with Galaxy release 2.2.3.

### DIFF
--- a/automation/roles/add_repository/tasks/main.yml
+++ b/automation/roles/add_repository/tasks/main.yml
@@ -33,6 +33,7 @@
   environment: "{{ proxy_env | default({}) }}"
   when: installation_method == "packages" and ansible_os_family == "Debian"
   tags: add_repo
+  check_mode: false
 
 - block: # RedHat/CentOS
     - name: Add repository GPG key
@@ -161,6 +162,7 @@
   environment: "{{ proxy_env | default({}) }}"
   when: installation_method == "packages" and ansible_os_family == "RedHat"
   tags: add_repo
+  check_mode: false
 
 - name: Extensions repository
   ansible.builtin.import_tasks: extensions.yml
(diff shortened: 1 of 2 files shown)


vitabaks:
@pat-s in this case, these tasks will make changes even in check mode, I don't think this is the expected behavior.

We could ignore the error if a check is being performed:
```yaml
ignore_errors: "{{ ansible_check_mode }}"
```

or not run the task in check mode:

```yaml
when: not ansible_check_mode
```

pat-s:
Using `ignore_errors` is a bad pattern IMO as it won't show you real errors in the first place which is actually why you are doing a dry_run. There is no other option than to actually execute some non-intrusive tasks to be able to check the possible correct execution of others.

An alternative is to skip all of these tasks in the first place but this has the same drawbacks as `ignore_errors`, i.e. you just don't check them at all. The goal is always to execute as many as possible so you can actually spot possible changes and errors in the config. Skipping should be the last resort.

I see why you would want to not execute some of them here as they would add a repo.
In the tasks of the diff one could be more granular and only run the ones which register a variable, so that this var is at least defined in the next step, preventing an error. And don't execute the ones which actually add the repo.

Your decision how to proceed, just wanted to bring up the point and issue :)

vitabaks:
as a user, I do not expect that actual changes will be made (e.q., creating a database).

pat-s:
This one will likely fail on a non-existing cluster anyhow yes, as the cluster doesn't exist. It worked on my existing one as the DBs and the cluster was already in place. So in this case, I'd say 

1. either add a conditional task which first checks if the DB/cluster exists and only then performs the task
2. Skip it entirely in check mode

pat-s:
@vitabaks I think you we're right, there's likely not a perfect solution for these cases and `ignore_errors` is not entirely avoidable. Have a look at the most recent changes.

PR #1134, opened 2025-07-12
fix: robustify private interface detection

pat-s:
The current logic simply takes the first private interface returned on a node as its address.
This might not always be the default private network interface but can also be a non-functional one or the default docker gateway.

This change makes the detection logic more robust and filters out common interface names which shouldn't be set as the default bind address.

As this task appears quite often across playbooks, it might be worth executing it (once) in a top-level role which always runs?


vitabaks:
Created the `bind_address` role so as not to duplicate the code.

PR #1171, opened 2025-07-27
fix etcd dir ownership

MParvin:
Problem:
When installing PostgreSQL clusters on Debian 12 and Oracle Linux 9, the Enable and start etcd service task fails because /etc/etcd is owned by root.

Error:
```
Jul 26 19:33:11 pg-cluster1 systemd[1]: etcd.service: Failed with result 'timeout'.
Jul 26 19:33:11 pg-cluster1 bash[22046]: {"level":"warn","ts":"...","msg":"failed to publish local member to cluster","error":"etcdserver: request timed out"}
```

Fix:
Added a task to set the correct owner (etcd) and permissions for /etc/etcd before starting the service.

Testing:
Verified on Oracle Linux 9—etcd now starts correctly.

### DIFF
--- a/automation/roles/etcd/tasks/main.yml
+++ b/automation/roles/etcd/tasks/main.yml
@@ -131,6 +131,15 @@
     dest: /etc/systemd/system/etcd.service
   tags: etcd, etcd_conf
 
+- name:  Ensurce proper permission /etc/etcd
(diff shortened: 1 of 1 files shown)


vitabaks:
@MParvin  Thanks for the contribution! I've made a small correction.

PR #1174, opened 2025-07-29
fix: `check_mode` improvements for `pgbackrest` role

pat-s:
similar to #1132 

without it, a dry-run of the config PB errors out when `pgbackrest` is included


vitabaks:
Thanks!
it will be necessary to think about how to add tests for the check mode in order to better support this.

PR #1198, opened 2025-08-14
fix: make `update_pgcluster` playbook work in Ansible `check_mode`

pat-s:
similar as #1132


vitabaks:
Thanks for the contribution!

check_mode isn’t officially supported yet — it may work for some playbooks but not all. 

I’ll work on adding test coverage and improving support for this mode.

PR #1146, opened 2025-07-15
Fix delegate to when fetch TLS certificates from master

evgeny-bushuev-incountry:

PR #1202, opened 2025-08-17
fix: various fixes for `add_pgnode` playbook

pat-s:
I faced various issues while executing this PB (with v2.3.1) due to undefined vars, issues when hosts are in multiple groups and the processing of raw command output in metadata filtering.

I am not sure if this was caused by some non-standard inventory setup, but AFAICS it should align with the suggested group patterns. With all these fixes, I was able to execute all tasks of the PB successfully. I am yet missing the `haproxy` LB on the new node and `patronitctl list` doesn't yet work, but this might be a different issue.

`ssh_keys.yml`:

Tasks were accessing SSH key variables from other hosts without checking if the variables were defined, causing 'dict object' has no attribute 'content' errors.

-> Replaced direct variable access with safer `.get()` method calls

The `known_hosts` module was receiving malformed SSH key data that included comment lines, causing "Host parameter does not match hashed host field" errors.

-> Filter out comment lines (starting with #) from ssh-keyscan output
-> Add safety checks to ensure valid SSH key data exists before processing

`copy.yml`:

The TLS certificate copy task was trying to access tls_files variable from the master node via hostvars, but the variable wasn't available when the `master` node wasn't included in the current play.

-> Use delegate_to to fetch TLS files directly from the master node
-> Set the `tls_files` fact locally instead of relying on hostvars access

---

My inventory looked like this:

```yaml
[etcd_cluster]
10.10.1.1 ansible_host=$IP 
10.10.1.2 ansible_host=$IP 
10.10.1.3 ansible_host=$IP 
10.10.5.2 ansible_host=$IP postgresql_exists=false new_node=true 

[balancers]
10.10.1.1 ansible_host=$IP  
10.10.1.2 ansible_host=$IP  
10.10.1.3 ansible_host=$IP 

[master]
10.10.1.1 hostname=pgnode1 ansible_host=$IP  postgresql_exists=false 
[replica]
10.10.1.2 hostname=pgnode2 ansible_host=$IP  postgresql_exists=false 
10.10.1.3 hostname=pgnode3 ansible_host=$IP  postgresql_exists=false 
10.10.5.2 ansible_host=$IP postgresql_exists=false new_node=true 

[pgbackrest]
10.11.1.1 ansible_host=$IP  
10.11.1.2 ansible_host=$IP 
10.11.1.3 ansible_host=$IP
10.10.5.2 ansible_host=$IP 

[postgres_cluster:children]
master
replica
```

### DIFF
--- a/automation/roles/tls_certificate/tasks/copy.yml
+++ b/automation/roles/tls_certificate/tasks/copy.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Fetch TLS certificate, key and CA from {{ inventory_hostname }}"
+- name: "Fetch TLS certificate, key and CA from master node"
   ansible.builtin.slurp:
     src: "{{ fetch_tls_dir | default(tls_dir | default('/etc/tls')) }}/{{ item }}"
   register: tls_files_raw
@@ -7,7 +7,8 @@
     - "{{ fetch_tls_privatekey | default(tls_privatekey | default('server.key')) }}"
     - "{{ fetch_tls_cert | default(tls_cert | default('server.crt')) }}"
     - "{{ fetch_tls_ca_cert | default(tls_ca_cert | default('ca.crt')) }}"
-  when: inventory_hostname == (groups[tls_group_name | default('master')][0])
+  delegate_to: "{{ groups[tls_group_name | default('master')][0] }}"
+  run_once: true
   tags: tls, tls_cert_copy
 
 - name: "Set variable: tls_files"
@@ -16,7 +17,7 @@
       server_key: "{{ tls_files_raw.results[0].content }}"
       server_crt: "{{ tls_files_raw.results[1].content }}"
       ca_crt: "{{ tls_files_raw.results[2].content }}"
-  when: inventory_hostname == (groups[tls_group_name | default('master')][0])
+  run_once: true
   tags: tls, tls_cert_copy
 
 - name: Create directory {{ copy_tls_dir | default(tls_dir | default('/etc/tls')) }}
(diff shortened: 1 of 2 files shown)


automation/roles/tls_certificate/tasks/copy.yml
```diff
@@ -1,13 +1,14 @@
...
+- name: "Fetch TLS certificate, key and CA from master node"
   ansible.builtin.slurp:
     src: "{{ fetch_tls_dir | default(tls_dir | default('/etc/tls')) }}/{{ item }}"
   register: tls_files_raw
   loop:
     - "{{ fetch_tls_privatekey | default(tls_privatekey | default('server.key')) }}"
     - "{{ fetch_tls_cert | default(tls_cert | default('server.crt')) }}"
     - "{{ fetch_tls_ca_cert | default(tls_ca_cert | default('ca.crt')) }}"
-  when: inventory_hostname == (groups[tls_group_name | default('master')][0])
+  delegate_to: "{{ groups[tls_group_name | default('master')][0] }}"
```
vitabaks, line 10: It seems that we recently got rid of delegate_to - https://github.com/vitabaks/autobase/pull/1146

vitabaks:
> I am yet missing the haproxy LB on the new node 

Use `add_balancer.yml` playbook. Doc: https://autobase.tech/docs/management/cluster-scaling (See "HAProxy" tab)

pat-s:
> Use add_balancer.yml playbook.

Jep, found that meanwhile.
Though I am not using it, as I need to integrate the nodes into a cluster which has a HAProxy running already. 
Instead, I am adding listeners to this HAProxy manually now.

vitabaks:
>  I am adding listeners to this HAProxy manually now.

As I understand it, we are talking about HAProxy, which is managed independently, not using Autobase. \
Since our haproxy scheme also uses confd, which will automatically update the configuration from etcd if the number of nodes in the patroni cluster has been changed.

automation/roles/tls_certificate/tasks/copy.yml
vitabaks: Yes, there is no way to do without `delegate_to`, since the add_pgnode playbook is [executed](https://github.com/vitabaks/autobase/blob/2.3.1/automation/playbooks/add_pgnode.yml#L139) in the **new_replica** (in-memory inventory) group.

This also means that we don't have enough tests for  cluster scaling scenario. I'll make this a separate PR.

vitabaks:
I also improved the ssh-keyscan task to fetch the SSH host key of the pgBackRest server (when a dedicated pgBackRest over SSH is used instead of S3) so that it is added to .ssh/known_hosts on the database servers.

Additionally, I fixed the template that generates the list of Postgres servers in the stanza section. Previously, it relied on inventory_hostname, which could cause errors when using an alias instead of an IP address in inventory file. Now it uses the value of the bind_address variable.

This is what I came across during testing.

PR #1204, opened 2025-08-17
feat: set node-size related config options dynamically per node

pat-s:
#1192 was a step into the right direction, however it still requires setting certain node settings dynamically, which ideally is handled by default through `autobase`.

This PR adds support for a dynamic generation of `shared_buffers` and `effective_cache_size`, by removing these from the DCS config and defining them instead in a node-specific PG config file. 

I've tested this approach in a setup with nodes having different memory sizes and verified that these have unique settings applied for the mentioned values through

```sh
psql -U postgres -c 'SHOW effective_cache_size'
psql -U postgres -c 'SHOW shared_buffers'
```


vitabaks:
This is not necessary. For host-level configuration, you can simply transfer these options to the `local_postgresql_parameters` variable. This will ensure that the options are stored in the patroni.yml file of each host, rather than in the DCS.

```yaml
local_postgresql_parameters:
  - { option: "shared_buffers", value: "{{ (ansible_memtotal_mb * 0.25) | int }}MB" } # by default, 25% of RAM
  - { option: "effective_cache_size", value: "{{ (ansible_memtotal_mb * 0.75) | int }}MB" } # by default, 75% of RAM
```

I did not do this by default, because I assumed that in most cases this is not necessary. It is recommended that all cluster nodes have the same configuration. But I think it's not a problem to store a couple of parameters locally to support such installations.

pat-s:
See my reasoning in the OP. I was aware of this when opening this PR. I don't think that the current model is a good long-term approach.

`autobase` already has so many parameters, both recognizing this one and setting it manually is a pain. On top, it is not needed and can be done dynamically behind the scenes.
As a user, I actually would expect that `autobase` is doing this for me. 

> It is recommended that all cluster nodes have the same configuration

I see the point and understand it, but recommendations are not always mirroring reality. Often enough, people have spare machines available from different sources that should serve as additional nodes. And in exactly these situations, I would expect `autobase` to set all relative parameters node-specific rather than inferring them from a central default.
I don't see any downside in this approach, as it won't change anything for active deployments. 

`local_postgresql_parameters` is still worthwhile for setting static node-specific parameters that should (for some reason) differ from the DCS default.

vitabaks:
> autobase already has so many parameters, both recognizing this one and setting it manually is a pain. 

Only advanced users who clearly understand what they are doing should work with parameters and command-line mode.
For beginners, we provide a UI that automatically prepares the variables, so the flexibility of Autobase and the large number of parameters is not a problem. The set of parameters will continue to grow in the future.

> On top, it is not needed and can be done dynamically behind the scenes.

This is not always the case — users should have the option to control such settings.
For example, one may prefer to set a specific value for shared_buffers instead of relying entirely on automation.


I’m not sure whether adding another configuration file (postgresql.node.conf) is actually more convenient than simply keeping custom parameters in patroni.yml. \
In the solution I propose, all parameters are defined in patroni.yml. They are then stored either in the DCS (for common parameters defined via the postgresql_parameters variable), or in the postgresql.parameters section of each host (to override the common ones) using the local_postgresql_parameters variable.

vitabaks:
@pat-s Please check this [commit](https://github.com/vitabaks/autobase/pull/1204/commits/a7af550676dd514ac73c4bc01b0d66cf29d2f6b4)

Doc: https://patroni.readthedocs.io/en/latest/patroni_configuration.html

pat-s:
> Only advanced users who clearly understand what they are doing should work with parameters and command-line mode.
For beginners, we provide a UI that automatically prepares the variables, so the flexibility of Autobase and the large number of parameters is not a problem. The set of parameters will continue to grow in the future.

I don't see how this fact relates to the fact being discussed. This point provides the option to make an important parameter (in fact 2) dynamic, so the resulting state has node-specific changes where needed. This is exactly what users expect from a tool focusing on *good* Postgres deployments. Exactly these settings shouldn't be outsourced to the user to find out the hard way or even stumble over it through a failing `patroni` restart in case `shared_buffers` is too large.

Also, side-discussion: it is somewhat tedious always having to discuss about the "UI vs advanced" mode and users when contributing (this came up a few times already). I think its fair to assume that pretty much all people contributing to the code here use automation in some form. UI users don't know/don't care about the specifics and just trust what is being shipped and also rarely want to tweak anything.

> For example, one may prefer to set a specific value for shared_buffers instead of relying entirely on automation.

As said, this is still possible by defining `local_postgresql_parameters` manually?

> lease check this [commit](https://github.com/vitabaks/autobase/pull/1204/commits/a7af550676dd514ac73c4bc01b0d66cf29d2f6b4)

Also fine if this works OOB without adding new config files. This would be even easier then, as there would be a central variable responsible for node-specific settings. Not sure about the naming though, as `local` isn't really expressive here (and only makes sense when knowing about DCS being the "remote" here). Otherwise one might also think "what does 'local' mean - I don't have a 'remote' node, do I?"

I guess something like `individual_node_override_postgresql_parameters` would be more descriptive, albeit being long.

vitabaks:
By default, the DCS is local, but it is often deployed externally (on dedicated servers), which is especially important in large infrastructures with heavy database loads. 

Perhaps the chosen variable name isn’t the most appropriate — I simply followed the Patroni [documentation](https://patroni.readthedocs.io/en/latest/patroni_configuration.html) in this case.

> Local configuration file patroni.yml

I was hoping that the new comments on the variables would clarify this.

vitabaks:
Example of configuration file:

`cat /etc/patroni/patroni.yml `

```yaml
---

scope: test-pgcluster
name: pgnode01
namespace: /service

log:
  level: WARNING
  traceback_level: ERROR
  format: '%(asctime)s %(levelname)s: %(message)s'
  dateformat: ''
  max_queue_size: 1000
  dir: /var/log/patroni
  file_num: 4
  file_size: 25000000
  loggers:
    patroni.postmaster: WARNING
    urllib3: WARNING

restapi:
  listen: 0.0.0.0:8008
  connect_address: 10.172.0.20:8008
  authentication:
      username: patroni
      password: O9b2qHntWdhOLsBeaTTxIEt3Jm0dXCSb
  request_queue_size: 5


consul:
  host: 127.0.0.1:8500
  checks: []
  scheme: https
  cacert: /etc/patroni/tls/consul/ca.crt
  cert: /etc/patroni/tls/consul/server.crt
  key: /etc/patroni/tls/consul/server.key

bootstrap:
  method: initdb
  dcs:
    ttl: 30
    loop_wait: 10
    retry_timeout: 10
    maximum_lag_on_failover: 1048576
    master_start_timeout: 300
    synchronous_mode: false
    synchronous_mode_strict: false
    synchronous_node_count: 1
    postgresql:
      use_pg_rewind: true
      use_slots: true
      parameters:
        max_connections: "1000"
        superuser_reserved_connections: "5"
        password_encryption: "scram-sha-256"
        ssl: "on"
        ssl_prefer_server_ciphers: "on"
        ssl_cert_file: "/etc/tls/server.crt"
        ssl_key_file: "/etc/tls/server.key"
        ssl_ca_file: "/etc/tls/ca.crt"
        ssl_min_protocol_version: "TLSv1.2"
        max_locks_per_transaction: "512"
        max_prepared_transactions: "0"
        huge_pages: "try"
        work_mem: "64MB"
        maintenance_work_mem: "512MB"
        checkpoint_timeout: "15min"
        checkpoint_completion_target: "0.9"
        min_wal_size: "1GB"
        max_wal_size: "8GB"
        wal_buffers: "32MB"
        wal_level: "logical"
        wal_keep_size: "1GB"
        wal_log_hints: "on"
        wal_compression: "on"
        bgwriter_delay: "20"
        bgwriter_lru_maxpages: "1000"
        default_statistics_target: "1000"
        seq_page_cost: "1"
        random_page_cost: "1.1"
        effective_io_concurrency: "200"
        synchronous_commit: "on"
        autovacuum: "on"
        autovacuum_max_workers: "5"
        autovacuum_analyze_scale_factor: "0.01"
        autovacuum_vacuum_scale_factor: "0.01"
        autovacuum_vacuum_insert_scale_factor: "0.1"
        autovacuum_vacuum_cost_limit: "500"
        autovacuum_vacuum_cost_delay: "2"
        autovacuum_naptime: "1s"
        archive_mode: "on"
        archive_timeout: "1800s"
        archive_command: "cd ."
        max_wal_senders: "20"
        max_replication_slots: "20"
        hot_standby: "on"
        hot_standby_feedback: "on"
        max_standby_streaming_delay: "30s"
        wal_receiver_status_interval: "10s"
        pg_stat_statements.max: "10000"
        pg_stat_statements.track: "all"
        pg_stat_statements.track_planning: "true"
        pg_stat_statements.track_utility: "false"
        pg_stat_statements.save: "true"
        auto_explain.log_min_duration: "10s"
        auto_explain.log_analyze: "true"
        auto_explain.log_buffers: "true"
        auto_explain.log_timing: "false"
        auto_explain.log_triggers: "true"
        auto_explain.log_verbose: "true"
        auto_explain.log_nested_statements: "true"
        auto_explain.sample_rate: "0.01"
        track_io_timing: "on"
        track_activities: "on"
        track_activity_query_size: "4096"
        track_counts: "on"
        track_functions: "all"
        log_lock_waits: "on"
        log_temp_files: "0"
        log_checkpoints: "on"
        log_rotation_age: "1d"
        log_rotation_size: "0"
        log_line_prefix: "%t [%p-%l] %r %q%u@%d "
        log_filename: "postgresql-%a.log"
        log_directory: "/var/log/postgresql"
        log_truncate_on_rotation: "on"
        logging_collector: "on"
        jit: "off"
        max_files_per_process: "4096"
        max_worker_processes: "16"
        max_parallel_workers: "8"
        max_parallel_workers_per_gather: "2"
        max_parallel_maintenance_workers: "2"
        tcp_keepalives_count: "10"
        tcp_keepalives_idle: "300"
        tcp_keepalives_interval: "30"
        idle_in_transaction_session_timeout: "10min"
        shared_preload_libraries: "pg_stat_statements,pg_stat_kcache,pg_wait_sampling,auto_explain,pg_cron,pgaudit"

  initdb:  # List options to be passed on to initdb
    - encoding: UTF8
    - locale: en_US.UTF-8
    - data-checksums

  pg_hba:  # Add following lines to pg_hba.conf after running 'initdb'
    - host replication replicator 127.0.0.1/32 scram-sha-256
    - host all all 0.0.0.0/0 scram-sha-256


postgresql:
  listen: 0.0.0.0:5432
  connect_address: 10.172.0.20:5432
  use_unix_socket: true
  data_dir: /pgdata/17/main
  bin_dir: /usr/lib/postgresql/17/bin
  config_dir: /etc/postgresql/17/main
  pgpass: /var/lib/postgresql/.pgpass_patroni
  authentication:
    replication:
      username: replicator
      password: MKhKK8FUsJHty99yvN0fdYkBJ6lBpnrP
    superuser:
      username: postgres
      password: 1TVEvglz3gYYe7w4eMneVs5TMvI5vjH7
  parameters:
    unix_socket_directories: /var/run/postgresql
    shared_buffers: "16074MB"
    effective_cache_size: "48222MB"


  remove_data_directory_on_rewind_failure: false
  remove_data_directory_on_diverged_timelines: false


  create_replica_methods:
    - basebackup
  basebackup:
    max-rate: '1000M'
    checkpoint: 'fast'


watchdog:
  mode: automatic  # Allowed values: off, automatic, required
  device: /dev/watchdog  # Path to the watchdog device
  safety_margin: 5

tags:
  datacenter: dc1
  key1: value1
  nosync: false
  noloadbalance: false
  nofailover: false
  clonefrom: false
```

Here we can see that the shared_buffers and effective_cache_size parameters are now defined under postgresql.parameters instead of bootstrap.dcs.postgresql.parameters.

This makes it possible to (atomaticaly) set individual values for each node based on the server’s available memory.

Issue #1222, opened 2025-08-25
[Bug] Patroni logging configuration can't be adjusted when patroni_log_destination: stderr


bkhomuts:
### Bug description

When config has:
`patroni_log_destination: stderr`
It is not possible to adjust logging parameters, such as the log level or log format.

As Patroni allows setting-up such parameters, even when logging to stderr, this limitation is caused by autobase, and can be fixed.

### Expected behavior

When patroni_log_destination: stderr, configuration of the `log` section shouldn't be skipped entirely by the autobase. Only settings related to files.

### Steps to reproduce

1. Set:
```
patroni_log_destination: stderr
patroni_log_format: "patroni: %(levelname)s: %(message)s"
patroni_log_dateformat: ""
```
2. Config (or deploy the cluster) with vitabaks.autobase.config playbook.
3. Check the /etc/patroni/patroni.yml configuration: the `log:` section is missing.

### Installation method

Command line

### System info

Autobase 2.3.0
The issue is not specific to OS.

### Additional info

Adjusting the log format is useful when logging with journald, and not to the individual files.

The issue can be fixed by updating the patroni.yml.j2 template, log section.

PR #1223, opened 2025-08-25
Allow adjusting logging parameters, even when patroni_log_destination is not logfile

bkhomuts:
Logging configuration is no longer skipped when logging to stderr.

Resolves: #1222


vitabaks:
OK, thanks for your contributing!

Issue #1228, opened 2025-08-27
[Feat] Allow specifying domain name in postgresql.connect_address


bkhomuts:
### Problem

At the moment, postgresql_connect address is defined as IP address, but it is useful to specify it as a domain, especially in case when ssl_verify=full is required: (#1227).

### Description

Patroni allows specifying domain name in postgresql.connect_address, which then will be used for replication. The ability to define it from autobase is essential for users, who would want to have superuser and replication connections to be operating in verify-full mode, where the server certificate will be verified also by the domain name.

### Importance

must have

### Proposed implementation

Quick implementation would be to define optional postgresql_connect_addr variable, which can be set to "{{ ansible_fqdn }}" for users, then updating  patroni.yml.j2 with:
connect_address: {{ postgresql_connect_addr | default(bind_address, true) }}:{{ postgresql_port }}

vitabaks:
We do not use domain addresses in the patroni configuration, there is no special need for this, and it is usually less reliable than using ip addresses.

vitabaks:
>  especially in case when ssl_verify=full is required

I think `require` or `verify-ca` is quite enough.

vitabaks:
> Quick implementation would be to define optional `postgresql_connect_addr` variable

It's not difficult. Feel free to suggest a PR.

```yaml
postgresql:
  listen: {{ postgresql_listen_addr }}:{{ postgresql_port }}
  connect_address: {{ postgresql_connect_address | default(patroni_bind_address | default(bind_address, true), true) }}:{{ postgresql_port }}
```

bkhomuts:
I need to have a domain name to ensure the certificate validation works with replication. If connection is done by the IP, then TLS certificate validation fails, due to CN (SAN) not matching the requested IP.

I'll propose the PR this week.

PR #1230, opened 2025-08-28
Allow customizing connect address and patroni postgresql authentication options

### LINKED ISSUE #1227: [Feat] Allow specifying additional parameters for postgresql.authentication superuser and replicator
### Problem

Important parameters, such as sslmode and sslcert for patroni superuser and replication user can't be set. As a result it is not possible to change the default sslmode from prefer to verify-full, and sslrootcert to system to have TLS server certificate validation.

### Description

Patroni allows setting[ additional parameters](https://patroni.readthedocs.io/en/latest/yaml_configuration.html#postgresql) for DB connections, such as sslmode and sslrootcert. But as those can't defined in autobase, patroni uses the default setting of sslmode=prefer, which may not be suitable for scenarios, where certificate verification is required.

Adding the ability to customize sslmode and sslrootcert allows to configure more secure TLS connections, protected from on-path attacks.

### Importance

must have

### Proposed implementation

Update roles/patroni/templates/patroni.yml.j2 to allow specifying custom settings for each user.

bkhomuts:
1. Introduce new variables `patroni_superuser_auth_options` and `patroni_replication_auth_options`. The new variables enable the option to harden TLS connection by requiring certificate verification.
Resolves #1227 .
2. Allow customizing connect_address with new variables `patroni_restapi_connect_addr` and `postgresql_connect_addr`. When set to "{{ hostname }}", postgres and patroni will connect using the domain name, instead of IP address. This allows certificate validation, when `sslmode=verify-full`, and TLS certificate is issued only for domain, but not for the IP.
Resolves: #1228

### DIFF
--- a/automation/roles/common/defaults/main.yml
+++ b/automation/roles/common/defaults/main.yml
@@ -15,16 +15,12 @@ patroni_cluster_name: "postgres-cluster" # the cluster name (must be unique for 
 patroni_superuser_username: "postgres"
 patroni_superuser_password: "" # Please specify a password. If not defined, will be generated automatically during deployment.
 patroni_superuser_auth_options:
-  - { option: "sslmode", value: "{{ 'require' if tls_cert_generate | bool else 'disable' }}" }
-  - { option: "sslkey", value: "{{ tls_dir }}/{{ tls_privatekey }}" }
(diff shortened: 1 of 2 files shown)


vitabaks:
@bkhomuts I made the sslmode set by default to ensure that the connection is encrypted. Then you can redefine the value at your discretion.

Issue #1231, opened 2025-08-28
[Feat] Allow setting use_unix_socket_repl: true


bkhomuts:
### Problem

When sslmode is set to verify-full for the replication user, patroni returns:
```
ERROR: Can not fetch local timeline and lsn from replication connection
Traceback (most recent call last):
File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 1091, in get_replica_timeline
  with self.get_replication_connection_cursor(**self.config.local_replication_address) as cur:
File "/usr/lib/python3.11/contextlib.py", line 137, in __enter__
  return next(self.gen)
         ^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/patroni/postgresql/__init__.py", line 1086, in get_replication_connection_cursor
  with get_connection_cursor(**conn_kwargs) as cur:
File "/usr/lib/python3.11/contextlib.py", line 137, in __enter__
  return next(self.gen)
         ^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/patroni/postgresql/connection.py", line 158, in get_connection_cursor
  conn = psycopg.connect(**kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/patroni/psycopg.py", line 136, in connect
  ret = _connect(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^
File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 122, in connect
  conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: connection to server at "localhost" (::1), port 5432 failed: server certificate for "postgresql.example.com" (and 1 other name) does not match host name "localhost"
```

### Description

Patroni connects to the machine using localhost, and not the socket, unless use_unix_socket_repl: true is set, resulting in tls validation failure when sslmode is verify-full. This is because patroni connects to the domain name "localhost", but the certificate has "postgresql.example.com" domain. To avoid this problem, it is necessary to set `use_unix_socket_repl: true` and add to pg_hba.conf the following:
`local	    replication		     replicator		                               scram-sha-256`
Then, patroni will connect to local machine using unix socket, avoiding TLS certificate validation.

### Importance

really want

### Proposed implementation

Update patroni.yml.j2 to allow setting use_unix_socket_repl: true, allow local replication connections from pg_hba.conf

PR #1236, opened 2025-09-01
Introduce patroni_use_unix_socket and patroni_use_unix_socket_repl

bkhomuts:
The new parameters allow configuring
postgresql.use_unix_socket
postgresql.use_unix_socket_repl

use_unix_socket_repl is particularly useful when TLS is enabled for replication user with sslmode=verify-full . Communicating over socket avoids a problem of validating TLS certificate for the domain localhost, to which patroni tries to connect by default, if use_unix_socket_repl isn't set. By connecting to the socket, encrypted connection isn't used, avoiding the validation problem.

Resolves: #1231

### DIFF
--- a/automation/roles/patroni/templates/patroni.yml.j2
+++ b/automation/roles/patroni/templates/patroni.yml.j2
@@ -150,8 +150,11 @@ bootstrap:
 postgresql:
   listen: {{ postgresql_listen_addr }}:{{ postgresql_port }}
   connect_address: {{ postgresql_connect_addr | default(patroni_bind_address | default(bind_address, true), true) }}:{{ postgresql_port }}
-{% if patroni_superuser_username == 'postgres' %}
+{% if patroni_superuser_username == 'postgres' or (patroni_use_unix_socket | default(false)) %}
   use_unix_socket: true
+{% endif %}
+{% if patroni_use_unix_socket_repl | default(false) %}
+  use_unix_socket_repl: true
 {% endif %}
   data_dir: {{ postgresql_data_dir }}
   bin_dir: {{ postgresql_bin_dir }}
--- a/automation/roles/patroni/templates/patroni.yml.j2
+++ b/automation/roles/patroni/templates/patroni.yml.j2
@@ -146,11 +146,10 @@ bootstrap:
     - host replication {{ patroni_replication_username }} 127.0.0.1/32 {{ postgresql_password_encryption_algorithm }}
     - host all all 0.0.0.0/0 {{ postgresql_password_encryption_algorithm }}
 
-
 postgresql:
   listen: {{ postgresql_listen_addr }}:{{ postgresql_port }}
   connect_address: {{ postgresql_connect_addr | default(patroni_bind_address | default(bind_address, true), true) }}:{{ postgresql_port }}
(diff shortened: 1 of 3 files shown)


automation/roles/patroni/templates/pg_hba.conf.j2
```diff
@@ -96,3 +96,6 @@
 {% for host in groups['postgres_cluster'] %}
   {% if tls_cert_generate | default(false) | bool %}hostssl{% else %}host{% endif %}      replication              {{ patroni_replication_username }}               {{ hostvars[host]['patroni_bind_address'] | default(hostvars[host]['bind_address'], true) }}/32         {{ postgresql_password_encryption_algorithm }}
 {% endfor %}
+{% if patroni_use_unix_socket_repl | default(false) %}
+  local      replication               {{ patroni_replication_username }}                                          {{ postgresql_password_encryption_algorithm }}
```
vitabaks: in this case, does it make sense to leave the replication rule for `localhost` ?

vitabaks:
cc @pat-s

automation/roles/patroni/templates/pg_hba.conf.j2
bkhomuts: Yes, it makes sense to remove it when socket is used for replication, as replicator no longer uses localhost --- confirmed with logs.

automation/roles/patroni/templates/patroni.yml.j2
```diff
@@ -150,7 +150,7 @@ bootstrap:
 postgresql:
   listen: {{ postgresql_listen_addr }}:{{ postgresql_port }}
   connect_address: {{ postgresql_connect_addr | default(patroni_bind_address | default(bind_address, true), true) }}:{{ postgresql_port }}
-{% if patroni_superuser_username == 'postgres' or (patroni_use_unix_socket | default(false)) %}
+{% if patroni_superuser_username == 'postgres' and patroni_use_unix_socket | default(false) %}
```
bkhomuts: In case we replace it with and, communicating over socket will be associated with postgres user, but there could be non-default user with superuser privileges, which was created specifically for patroni.
So I's suggest to return it back to 'or'.

vitabaks: I simplified the condition

vitabaks:
https://github.com/vitabaks/autobase/pull/1236/commits/df3060550a1a9deed4d81a233c2750c3820a9627

before:

```
# Allow replication connections from localhost, by a user with the
# replication privilege.
  local      replication               replicator                                          trust
  hostssl      replication              replicator               10.172.0.20/32         scram-sha-256
  hostssl      replication              replicator               10.172.0.21/32         scram-sha-256
  hostssl      replication              replicator               10.172.0.22/32         scram-sha-256
```

after:

```
# Allow replication connections from localhost, by a user with the
# replication privilege.
  local        replication              replicator                                      trust
  hostssl      replication              replicator               10.172.0.20/32         scram-sha-256
  hostssl      replication              replicator               10.172.0.21/32         scram-sha-256
  hostssl      replication              replicator               10.172.0.22/32         scram-sha-256
```

Issue #1280, opened 2025-09-30
Directory permissions for /var/log/postgresql are changed by Autobase


bkhomuts:
### Bug description

Autobase resets permissions of `/var/log/postgresql` during reconfiguration. The permissions are different than default on Debian, which creates issues for the logging software, if it needs to list `/var/log/postgresql`.

### Expected behavior

Only the needed permissions of the `/var/log/postgresql` directory are changed.

### Steps to reproduce

1. chmod 755 `/var/log/postgresql`
2. Run vitabaks.autobase.config_pgcluster
3. Read-execute on the directory is removed from others.

### Installation method

Command line

### System info

Autobase v2.4.0

### Additional info

The mode for the log directory is specified as 0750 and 0700. If it is changed to `u=rwx,g=rx` and `u=rwx`, this will result in group and other permissions remain without change.

PR #1281, opened 2025-09-30
Don't change permissions for the log directory when not needed

bkhomuts:
Don't change permissions for the log directory when not needed

This change ensures that group and other permissions on the log 
directory aren't changed by autobase, when not needed.

Resolves: #1280

### DIFF
--- a/automation/roles/patroni/tasks/config.yml
+++ b/automation/roles/patroni/tasks/config.yml
@@ -15,7 +15,7 @@
     owner: postgres
     group: postgres
     state: directory
-    mode: "0750"
+    mode: "u+rwx,g+rx"
   when: patroni_log_destination == 'logfile'
   tags: patroni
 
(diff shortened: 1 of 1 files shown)


vitabaks:
Why switch from octal (0750, 0700) to symbolic (u=rwx,g=rx, u=rwx) if the permissions are effectively the same?

bkhomuts:
> Why switch from octal (0750, 0700) to symbolic (u=rwx,g=rx, u=rwx) if the permissions are effectively the same?

The difference is that group and other permissions are not reset. In the octal version, it removes already existing permissions - changes it from the default or from the setting of the user.

The new version only adds the missing permissions, but does not remove them, if the administrator set a different value. Particularly useful for syslog-ng, which needs to list files in the directory when wildcard-file source is used.

PR #1288, opened 2025-10-01
fix: Make sure PostgreSQL unix socket, log and home directory exists

cherts:
Hi,

while testing autobase, I encountered errors when the Postgres home, logs, or unix socket directory didn't exist or had incorrect permissions and ownership.

These are very rare situations, but I simulated them, and the autobase not correct working.

Therefore, we must check and create these directories in any case.


vitabaks:
Ok, thanks

PR #1290, opened 2025-10-02
Run scheduled workflows only repo owner vitabaks

cherts:
Run scheduled workflows only repo owner vitabaks

I don't think scheduled tasks should be done in forks

Let's not heat our planet unnecessarily. ;)


vitabaks:
good point, thanks!

:) 

image

PR #1292, opened 2025-10-02
fix: Fixed several errors in the remove_node playbook

cherts:
Hi

I tested the working of remove_node.yml playbook and found a many of errors:

1) Deleting etcd nodes didn't work because bind_address was defined only at the postgres_cluster host group level.
2) The "Update patroni config file" step overwritten all passwords in the patroni.yml file because vitabaks.autobase.pre_checks wasn't called.
3) The "Read postgres public SSH key from target node" step had an error because the `~postgres/.ssh` directory might not exist. The error was ignored, but it generated unnecessary red messages.

I fixed all three of these errors. Now deleting the postgres node works correctly.

New feature:
1) Remove a node from pg_hba.conf on existing cluster nodes

However, I noticed that deleting doesn't regenerate the SSL certificates for postgres; they will contain the address and name of the deleted node. This isn't critical, and I'll try to fix it in a next MR.

Thank's

### DIFF
--- a/automation/playbooks/remove_node.yml
+++ b/automation/playbooks/remove_node.yml
@@ -354,7 +354,7 @@
       tags: postgres, postgresql, etcd
 
     - block:
-        - name: Waiting for 10 seconds after Patroni cluster reload
+        - name: Waiting for 20 seconds after Patroni cluster reload
           ansible.builtin.pause:
             seconds: 20
 
--- a/automation/playbooks/remove_node.yml
+++ b/automation/playbooks/remove_node.yml
@@ -154,13 +154,6 @@
   vars:
     target_node: "{{ node_to_remove | default('') }}"
 
-  pre_tasks:
-    - name: Expose etcd_bind_address as facts
-      ansible.builtin.set_fact:
-        etcd_bind_address: "{{ etcd_bind_address | default(bind_address, true) }}"
-      when: dcs_type | default('etcd') == 'etcd'
-      tags: always
-
   tasks:
     - block:
         - name: Fetch etcd cluster members before removal
--- a/automation/playbooks/remove_node.yml
+++ b/automation/playbooks/remove_node.yml
@@ -356,7 +356,7 @@
     - block:
         - name: Waiting for 10 seconds after Patroni cluster reload
           ansible.builtin.pause:
-            seconds: 10
+            seconds: 20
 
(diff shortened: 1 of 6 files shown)


vitabaks:
It's strange, we conducted tests before the release, and everything was fine.

I don't think it's necessary to re-generate certificates when removing a node.

cherts:
I'm testing with etcd running on dedicated servers. Perhaps this somehow affected your tests.

vitabaks:
We definitely lack testers. But they will definitely be there with the growth of the project.

Thanks for your help!

cherts:
I can deploy virtual machines and test when etcd and postgres are running together.

I haven't yet tested adding and removing new etcd nodes — that's the next step.

vitabaks:
@cherts please let me know when you are completely done working on this PR.

cherts:
@vitabaks hi

I've completed testing the playbook remove_node.yml

1) I tested a cluster with etcd and postgres running on the same server (3 node + 1 node add/remove).
2) I tested a cluster with etcd and postgres running on separate servers (3 node etcd + 3 node postgres + 1 etcd node add/remove + 1 postgres node add/remove).

The fixed remove_node.yml playbook works correctly in all cases.

I also found an error in the add_node.yml playbook, but that will be a new MR to fix.

Thank's

automation/playbooks/remove_node.yml
```diff
@@ -1,4 +1,29 @@
...
+        name: vitabaks.autobase.bind_address
+      tags: always
+
+    - name: Set maintenance variable
+      ansible.builtin.set_fact:
+        postgresql_cluster_maintenance: true
+      tags: always
+
+  roles:
+    - role: vitabaks.autobase.pre_checks
```
vitabaks: @cherts Why do we need pre-checks for deletion?

automation/playbooks/remove_node.yml
```diff
@@ -328,7 +364,22 @@
         - inventory_hostname != target_node
       tags: postgres, postgresql, etcd
 
+    - name: Remove a node from pg_hba.conf on existing cluster nodes
```
vitabaks, line 356: Is this really necessary? After all, if you delete the server from inventory, it will still be deleted from pg_hba

vitabaks:
@cherts I've made a few minor edits.

PR #1294, opened 2025-10-03
fix: Added update patroni config file in the add_node playbook

cherts:
Problem: After adding new etcd nodes, they are not added to the patroni configuration.

This MR solves this problem.


automation/playbooks/add_node.yml
```diff
@@ -343,3 +344,19 @@
 
     # finish (info)
     - role: vitabaks.autobase.deploy_finish
```
vitabaks: please move the deploy_finish role to the end

Issue #1256, opened 2025-09-16
[Feat] Check syntax of config files before reload or restart


bkhomuts:
### Problem

At the moment, it is possible to introduce an error in the postgresql configuration files, which will remain unnoticed until PostgreSQL server restart.
As Patroni is used for high availability, it is important to prevent errors, which may potentially lead to downtime.

### Description

Introduce syntax check of the configuration file at least before reloading services (or better before updating the config files). This will warn the administrator if an error in configuration has been made.

### Importance

nice to have

vitabaks:
Perhaps `patroni --validate-config` is suitable here.

https://patroni.readthedocs.io/en/latest/patroni_configuration.html#validate-patroni-configuration

bkhomuts:
Yes, it is useful. 

> Perhaps `patroni --validate-config` is suitable here.
> 
> https://patroni.readthedocs.io/en/latest/patroni_configuration.html#validate-patroni-configuration

And perhaps also checking the result of PostgreSQL reload, and reporting if there is an error.

PR #1297, opened 2025-10-06
Check syntax of patroni and pg_hba config files

sshenderson:
#1256

### DIFF
--- a/automation/molecule/tests/postgres/replication.yml
+++ b/automation/molecule/tests/postgres/replication.yml
@@ -1,11 +1,10 @@
 ---
 - name: Check PostgreSQL replication status
-  community.postgresql.postgresql_query:
-    login_host: "127.0.0.1"
+  postgresql_query:
+    query: "SELECT pg_is_in_recovery(), count(*) FROM pg_stat_wal_receiver;"
+    login_unix_socket: "{{ postgresql_unix_socket_dir }}"
     login_port: "{{ postgresql_port }}"
     login_user: "{{ patroni_superuser_username }}"
-    login_password: "{{ patroni_superuser_password }}"
-    login_db: "postgres"
-    query: "SELECT pg_is_in_recovery(), count(*) FROM pg_stat_wal_receiver;"
+    login_db: template1
   register: pg_replication_status
   failed_when: "pg_replication_status.query_result[0].pg_is_in_recovery and pg_replication_status.query_result[0].count == 0"


automation/roles/patroni/templates/patroni.yml.j2
```diff
@@ -69,7 +69,6 @@ etcd3:
 {% if dcs_type == 'consul' %}
 consul:
   host: 127.0.0.1:8500
-  checks: []
```
vitabaks, line 72: This change is not related to the PR topic

@sshenderson could you describe in detail why you deleted the option?

vitabaks: > By default an empty list is used.
https://patroni.readthedocs.io/en/latest/yaml_configuration.html#consul

an empty list has already been added by default, perhaps now it is really not necessary to explicitly specify this option.

vitabaks:
cc @bkhomuts

automation/roles/patroni/templates/patroni.yml.j2
sshenderson: @vitabaks  `--validate-config ` enforces that all lists in the configuration must be non-empty. If a list is empty, the validation fails:
https://github.com/patroni/patroni/blob/26de206aca13d75b874f0cdfba24c0561c908094/patroni/validator.py#L735-L754

automation/roles/patroni/tasks/main.yml
```diff
@@ -534,6 +534,18 @@
             group: postgres
             mode: "0640"
 
+        - name: Prepare PostgreSQL | validate generated pg_hba.conf
+          become: true
+          become_user: postgres
+          community.postgresql.postgresql_query:
+            login_unix_socket: "{{ postgresql_unix_socket_dir }}"
```
bkhomuts: This change assumes UNIX socket is used, but there is an option to use an IP address instead.

bkhomuts:
I have noticed that there is an assumption that UNIX socket is used, but autobase allows setting up patroni_use_unix_socket, which may be false.

Based on the comments, I suggest improving this PR:
1. Check if patroni_use_unix_socket is used, and connect either to IP or the socket.
2. Specify password for the superuser, as trust-based authentication may be disabled.

automation/roles/patroni/tasks/main.yml
```diff
@@ -534,6 +534,18 @@
...
+            login_port: "{{ postgresql_port }}"
+            login_user: "{{ patroni_superuser_username }}"
+            login_db: postgres
+            query: "SELECT * FROM pg_hba_file_rules WHERE error IS NOT NULL"
```
bkhomuts, line 546: ... and an assumption that postgres user doesn't require password.

vitabaks:
Just use 127.0.0.1 and PGPASSWORD env.

bkhomuts:
> Just use 127.0.0.1 and PGPASSWORD env.

+1 for consistency with other places where autobase makes PostgreSQL queries.

bkhomuts:
Note, that this change only validates pg_hba.conf, but it doesn't validate PostgreSQL server configuration.

vitabaks:
@bkhomuts 
image

bkhomuts:
@vitabaks ,
Thanks for pointing this out. I have meant validation of **PostgreSQL** configuration, specifically when incompatible or wrong parameters have been set. Is it covered also by patroni --validate-config ?

vitabaks:
We need to check, I believe that validate-config should be enough since it is patroni that manages the configuration.

Issue #1310, opened 2025-10-17
[Bug] Patroni ports already in use


pat-s:
### Bug description

Tried again applying latest state on my dev cluster with the following result:

(`pg_config` playbook)

```
TASK [vitabaks.autobase.patroni : Generate /etc/patroni/patroni.yml] **************************************************
[ERROR]: Task failed: Module failed: failed to validate
Origin: /$PATH/ansible_collections/vitabaks/autobase/roles/patroni/tasks/patroni.yml:2:3

1 ---
2 - name: "Generate {{ patroni_config_file | default('/etc/patroni/patroni.yml') }}"
    ^ column 3

fatal: [10.11.1.1]: FAILED! => {"changed": false, "checksum": "6c4e376aba88c153367b6534fa9ab93782319455", "exit_status": 1, "msg": "failed to validate", "stderr": "restapi.listen 0.0.0.0:8008 didn't pass validation: Port 8008 is already in use.\npostgresql.listen 0.0.0.0:5432 didn't pass validation: Port 5432 is already in use.\n", "stderr_lines": ["restapi.listen 0.0.0.0:8008 didn't pass validation: Port 8008 is already in use.", "postgresql.listen 0.0.0.0:5432 didn't pass validation: Port 5432 is already in use."], "stdout": "", "stdout_lines": []}
fatal: [10.11.1.3]: FAILED! => {"changed": false, "checksum": "e50ddfcc5857a839d5338ad18f8a8f9658955db3", "exit_status": 1, "msg": "failed to validate", "stderr": "restapi.listen 0.0.0.0:8008 didn't pass validation: Port 8008 is already in use.\npostgresql.listen 0.0.0.0:5432 didn't pass validation: Port 5432 is already in use.\n", "stderr_lines": ["restapi.listen 0.0.0.0:8008 didn't pass validation: Port 8008 is already in use.", "postgresql.listen 0.0.0.0:5432 didn't pass validation: Port 5432 is already in use."], "stdout": "", "stdout_lines": []}
[...]

PR #1317, opened 2025-10-22
Add ignore-listen-port option for Patroni config validation

sshenderson:
#1310


vitabaks:
**Important note**! This means that now we do not support Patroni versions older than [4.0.0](https://github.com/patroni/patroni/blob/master/docs/releases.rst#version-400) (Released 2024-08-29), since it was in this version that support for the`--ignore-listen-port` argument appeared.

PR #1320, opened 2025-10-24
fix: honor `pgbackrest_archive_command` for `archive_command` in `patroni.yaml`

pat-s:
This is currently broken (and according to my backup history since ~ 2 months).

With the following config

```yaml
pgbackrest_install: true
pgbackrest_repo_type: 's3'
pgbackrest_archive_command: 'pgbackrest --stanza={{ pgbackrest_stanza }} archive-push %p'
[...]
```

the `bootstrap.dcs.postgresql.parameters.archive_command` option in `patroni.yml` was still set to the default `cd .`. This caused all `pgbackrest` cron runs to fail.

Maybe there is a better way to fix this, but the following patch works.

### DIFF
--- a/automation/roles/patroni/tasks/config.yml
+++ b/automation/roles/patroni/tasks/config.yml
@@ -1,5 +1,34 @@
 ---
 # Patroni configure
+
+# Ensure archive_command is set correctly for pgbackrest (same logic as pre_checks/pgbackrest.yml)
+- name: "pgBackRest | Ensure 'archive_command' is set to '{{ pgbackrest_archive_command | default('') }}'"
+  ansible.builtin.set_fact:
+    postgresql_parameters: >-
+      {{ postgresql_parameters | rejectattr('option', 'equalto', 'archive_command') | list
+      + [{'option': 'archive_command', 'value': pgbackrest_archive_command}] }}
+  vars:
+    archive_command_item: "{{ postgresql_parameters | selectattr('option', 'equalto', 'archive_command') | list | last }}"
+  when:
+    - pgbackrest_install is defined
+    - pgbackrest_install | bool
+    - archive_command_item is undefined or archive_command_item.value != pgbackrest_archive_command
+  tags: patroni, patroni_conf
+
+# Ensure archive_command is set correctly for wal-g (same logic as pre_checks/wal_g.yml)
+- name: "WAL-G | Ensure 'archive_command' is set to '{{ wal_g_archive_command | default('') }}'"
+  ansible.builtin.set_fact:
+    postgresql_parameters: >-
+      {{ postgresql_parameters | rejectattr('option', 'equalto', 'archive_command') | list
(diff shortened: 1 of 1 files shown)


vitabaks:
it is assumed that the user will specify the desired archive_command in the postgresql_parameters variable, example:

```yaml
postgresql_parameters:
  - { option: "archive_command", value: "{{ pgbackrest_archive_command }}" }
  - ...
```

But if he didn't do it explicitly, then we substitute him at the pre-checks level - https://github.com/vitabaks/autobase/blob/2.4.1/automation/roles/pre_checks/tasks/pgbackrest.yml

automation/roles/patroni/tasks/config.yml
```diff
@@ -1,5 +1,34 @@
 ---
 # Patroni configure
+
+# Ensure archive_command is set correctly for pgbackrest (same logic as pre_checks/pgbackrest.yml)
```
vitabaks, line 4: I don't quite understand why we should duplicate the logic here, because we are already set archive_command in pre_checks role.

pat-s:
> it is assumed that the user will specify the desired archive_command in the postgresql_parameters variable

Ah! Though I am not sure if this is best approach:

- Doing so is non-intuitive, especially if there are dedicated `pgbackrest_*` vars available including the `pgbackrest_archive_command`
- `pgbackrest_archive_command` seems to be a no-op right now. If it shouldn't be used at all, it should be removed.

I personally try to avoid setting `postgresql_parameters` as it requires adding the full default definition into my group_vars, as otherwise these params aren't set anymore. However, this also means I can't easily spot anymore which of these I've now actually changed and which ones are defaults. This is a known issue with (long) variable list definitions in Ansible.

From a user PoV, I'd highly prefer `pgbackrest_archive_command` being respected (if set). This is what this PR is doing. 
I see it duplicates code in some ways but in the end it just ensure a proper use of the var in the first place. Maybe there are better ways but I don't think the current way is a good approach in the longterm.

vitabaks:
I'm thinking of adding a `postgresql_archive_command` variable that will dynamically substitute a value based on which backup tool is being used, and use this variable as the value of the archive_command option.

pat-s:
That might be an option yes. From a user PoV, I'd be happy not having to set `postgresql_archive_command` or `pgbackrest_archive_command` unless I really want to put something modified there.

The best would be to have a default for backups in place and even when deciding for another tool, only having to use `<tool name>_enabled: true` or similar to use it.

pat-s:
@vitabaks Are there any plans to move forward here or find a different solution? This issue requires me to run on a branch as otherwise my backups aren't functional.

vitabaks:
Yes, I'll do it.

vitabaks:
https://github.com/vitabaks/autobase/pull/1362

@pat-s please check.

vitabaks:
Done.

PR #1323, opened 2025-10-27
fix: Replaced tls_cert_generate with host_type in pg_hba.conf.j2

catemiko:
Replaced `tls_cert_generate` with `host_type` to honor partly disabled SSL.

### DIFF
--- a/automation/roles/patroni/templates/pg_hba.conf.j2
+++ b/automation/roles/patroni/templates/pg_hba.conf.j2
@@ -85,18 +85,18 @@
 
 # TYPE      DATABASE                 USER                     ADDRESS                  METHOD
 {% for client in postgresql_pg_hba %}
-  {{ client.type.ljust(9) |default('{% if tls_cert_generate | default(false) | bool %}hostssl{% else %}host{% endif %}') }} {{ client.database.ljust(24) |default('all') }} {{ client.user.ljust(24) |default('all') }} {{ client.address.ljust(24) |default('') }} {{ client.method |default('md5') }} {{ client.options |default(None) }}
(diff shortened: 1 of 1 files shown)


automation/roles/patroni/templates/pg_hba.conf.j2
```diff
@@ -85,18 +85,18 @@
 
 # TYPE      DATABASE                 USER                     ADDRESS                  METHOD
 {% for client in postgresql_pg_hba %}
-  {{ client.type.ljust(9) |default('{% if tls_cert_generate | default(false) | bool %}hostssl{% else %}host{% endif %}') }} {{ client.database.ljust(24) |default('all') }} {{ client.user.ljust(24) |default('all') }} {{ client.address.ljust(24) |default('') }} {{ client.method |default('md5') }} {{ client.options |default(None) }}
+  {{ client.type.ljust(9) |default('{{ host_type }}') }} {{ client.database.ljust(24) |default('all') }} {{ client.user.ljust(24) |default('all') }} {{ client.address.ljust(24) |default('') }} {{ client.method |default('md5') }} {{ client.options |default(None) }}
```
vitabaks, line 88: This is the default value, which takes effect only if `type` is not specified. To partially disable TLS, it is enough to specify "host" in type for the `postgresql_pg_hba` variable.

Example:

```yaml

postgresql_pg_hba:
...
  - { type: "host", database: "mydatabase", user: "mydb-user", address: "192.168.0.74/32", method: "{{ postgresql_password_encryption_algorithm }}" }
```

I'm not sure if this change is necessary.

catemiko: I encountered replicas health check error due to `hostssl` in `pg_hba.conf`, because i disabled SSL for PostgreSQL. Just to notice, `postgresql_pg_hba` can be used only for manually defined entries.

vitabaks: It's clear now. 

You need to be able to disable TLS in Postgres while still leaving it in other cluster components.

catemiko: Exactly!

PR #1360, opened 2025-11-25
fix: postgres CRON jobs for pgbackrest on AlmaLinux

pat-s:
I noticed that the `pgbackrest` cron jobs were not running and investigated.
Turns out that the current setup isn't working the way its configured right now:

On RHEL, the ``postgres`` user is created with a locked password (! in `/etc/shadow)`). While this prevents login, `pam_unix.so` also blocks cron execution for locked accounts. Changing to `*` (unusable password) allows cron while still preventing direct login.

### DIFF
--- a/automation/roles/cron/tasks/main.yml
+++ b/automation/roles/cron/tasks/main.yml
@@ -52,14 +52,7 @@
         name: postgres
         password: '*'
         password_lock: false
-      vars:
-        postgres_shadow_password: >-
-          {{
-            (postgres_shadow.stdout | default(''))
-            | regex_search('^[^:]*:([^:]+):', '\1')
-            | default('')
-          }}
-      when: postgres_shadow_password is match('^!+')
+      when: postgres_shadow.stdout is match('^postgres:!+')
   when:
     - ansible_os_family == "RedHat"
     - cron_jobs is defined and cron_jobs | length > 0


vitabaks:
What if the user decides to use a postgres user with password authorization? It is unlikely, but it should not be completely ruled out. 

Perhaps we need to check that the user really has a locked password before of specifying a password `*`.

vitabaks:
Additionally refactored cron configuration: removed duplicated cron tasks from the pgbackrest and wal_g roles and unified them via the shared cron role.

vitabaks:
Tested

```
TASK [Add pgbackrest cron jobs on database server] *****************************
included: vitabaks.autobase.cron for 10.116.0.17
TASK [vitabaks.autobase.cron : Check password lock state for postgres user] ****
ok: [10.116.0.17]
TASK [vitabaks.autobase.cron : Unlock the locked password for postgres user (for PAM/cron compatibility)] ***
changed: [10.116.0.17]
TASK [vitabaks.autobase.cron : Add cron jobs] **********************************
changed: [10.116.0.17] => (item={'name': 'pgBackRest: Full Backup', 'file': '/etc/cron.d/pgbackrest-postgres-cluster-01', 'user': 'postgres', 'minute': '00', 'hour': 1, 'day': '*', 'month': '*', 'weekday': '0', 'job': 'pgbackrest --stanza=postgres-cluster-01 --type=full backup'})
changed: [10.116.0.17] => (item={'name': 'pgBackRest: Diff Backup', 'file': '/etc/cron.d/pgbackrest-postgres-cluster-01', 'user': 'postgres', 'minute': '00', 'hour': 1, 'day': '*', 'month': '*', 'weekday': '1-6', 'job': 'pgbackrest --stanza=postgres-cluster-01 --type=diff backup'})
```

result:
```
[root@postgres-cluster-01-pgnode01 ~]# getent shadow postgres
postgres:*:20422::::::
```

PR #1362, opened 2025-11-26
Add dynamic PostgreSQL archive_command variable

vitabaks:
Replaces hardcoded archive_command with a new `postgresql_archive_command` variable that selects the appropriate command based on installed tools (pgBackRest, WAL-G, or disables with /bin/true).


pat-s:
Thanks! Looks good on the first look, will try it out when a new release is available!

Issue #1382, opened 2025-12-09
[Feat] Add pgBackRest bootstrap configuration directly to patroni.yml


bkhomuts:
### Problem

Patroni allows bootstrapping PostgreSQL clusters with pgbackrest. But autobase doesn't allow configuring this under the `bootstrap:` section.

### Description

Add an option to configure bootstrap section in the patroni.yml configuration file.

This would allow not only to create replicas with pgbackrest, but to bootstrap the cluster. Particularly useful when creating a copy of the existing cluster for testing.

### Importance

really want

### Proposed implementation

Update patroni.yml.j2 template, and add sections similar to `wal-g` and `pg_probackup`.

vitabaks:
Autobase has been supporting the creation of clusters from pgbackrest backups for many years.

Follow the documentation: https://autobase.tech/docs/management/restore

bkhomuts:
Hello @vitabaks ,
Thanks for linking the documentation, I have tried it today, but it wasn't sufficient. This is because the configuration of the bootstrap method pgbackrest was missing. In the autobase template, the config is as follows:
```
  method: pgbackrest
```
But for bootstrap to work, extra configuration is required, such as:
```
  pgbackrest:
    command: /usr/bin/pgbackrest --stanza=postgres-cluster --log-timestamp=n --delta restore
    keep_existing_recovery_conf: False
    no_params: True
    recovery_conf:
      recovery_target_action: promote
      recovery_target_timeline: latest
      restore_command: pgbackrest --stanza=postgres-cluster archive-get %f "%p"
```

vitabaks:
We don't use patroni for this, autobase will restore from the backup and then just run patroni for the ready-made database. 
https://github.com/vitabaks/autobase/blob/2.5.0/automation/roles/patroni/tasks/main.yml#L683-L790

This is necessary to support [delta restore](https://pgbackrest.org/user-guide.html#restore/option-delta)

bkhomuts:
@vitabaks , very much appreciate for pointing me to code. I haven't run the full playbook, that's why the cluster restore wasn't running restore_command.

But may I add an extra config section for bootstrap with pgBackRest to work from Patroni itself? It shouldn't create issues but would be valuable for my use in the dev environment, where I reprovision the patroni cluster myself, without autobase.

The idea is: Autobase installs the config files, but the cluster can be provisioned from pgBackRest without autobase.

vitabaks:
[...]

PR #1384, opened 2025-12-11
Add pgBackRest bootstrap configuration directly to patroni.yml

bkhomuts:
When bootstrap method is pgbackrest, a corresponding pgbackrest configuration section will be added under bootstrap in patroni.yml. This change allows bootstrapping the cluster using pgBackRest with Patroni, not relying on Autobase.

Resolves: #1382

### DIFF
--- a/automation/roles/patroni/tasks/main.yml
+++ b/automation/roles/patroni/tasks/main.yml
@@ -381,12 +381,6 @@
   when: postgresql_wal_dir is defined and postgresql_wal_dir | length > 0
   tags: patroni, custom_wal_dir
 
-- name: Check if data directory exists, before creating it
-  ansible.builtin.stat:
-    path: "{{ postgresql_data_dir }}"
-  register: stat_postgresql_data_dir
-  tags: patroni, point_in_time_recovery
-
 - block: # when postgresql NOT exists or PITR
     - name: Prepare PostgreSQL | make sure PostgreSQL home directory "{{ postgresql_home_dir | default('') }}"
       ansible.builtin.file:
@@ -470,6 +464,17 @@
             group: postgres
             mode: "0755"
 
+        - name: Prepare PostgreSQL | make sure the data directory "{{ postgresql_data_dir | default('') }}" is empty
+          ansible.builtin.file:
+            path: "{{ postgresql_data_dir }}"
+            state: "{{ item }}"
+            owner: postgres
+            group: postgres
+            mode: "0700"
+          loop:
+            - absent
+            - directory
+
         - name: Prepare PostgreSQL | make sure PostgreSQL unix socket directory "{{ postgresql_unix_socket_dir | default('') }}" exists
           ansible.builtin.file:
             path: "{{ postgresql_unix_socket_dir }}"
@@ -499,20 +504,6 @@
           when: postgresql_wal_dir | default('') | length > 0
       when: patroni_cluster_bootstrap_method != "pgbackrest" or pgbackrest_patroni_cluster_clean_bootstrap | default(false)
   when: not postgresql_exists | default(false) | bool or patroni_cluster_bootstrap_method != "initdb"
-  tags: patroni, point_in_time_recovery
-
-- name: Prepare PostgreSQL | make sure the data directory "{{ postgresql_data_dir | default('') }}" is empty
-  ansible.builtin.file:
-    path: "{{ postgresql_data_dir }}"
-    state: "{{ item }}"
-    owner: postgres
-    group: postgres
-    mode: "0700"
-  loop:
-    - absent
-    - directory
-  # Clean the data directory if it didn't exist before.
-  when: (not stat_postgresql_data_dir.stat.exists) or pgbackrest_patroni_cluster_clean_bootstrap | default(false)
   tags: patroni, point_in_time_recovery
 
 - block: # when postgresql exists


vitabaks:
Please perform the test and attach a log confirming that it works. 

Also please check that this does not break the current cluster recovery logic according to the [documentation](https://autobase.tech/docs/management/restore#command-line) (in theory, there should be no problems, but it is advisable to attach a log).

automation/roles/patroni/templates/patroni.yml.j2
```diff
@@ -90,6 +90,18 @@ bootstrap:
       {% endfor %}
       {% endfor %}
 {% endif %}
+{% if patroni_cluster_bootstrap_method == 'pgbackrest' %}
+  pgbackrest:
+    command: "{{ pgbackrest_patroni_cluster_restore_command }}"
+    keep_existing_recovery_conf: False
```
vitabaks: Perhaps the value should be `True`?

> If keep_existing_recovery_conf is defined and set to True, Patroni will not remove the existing recovery.conf file if it exists (PostgreSQL <= 11). Similarly, in that case Patroni will not remove the existing recovery.signal or standby.signal if either exists, nor will it override the configured recovery settings (PostgreSQL >= 12). This is useful when bootstrapping from a backup with tools like **pgBackRest** that generate the appropriate recovery configuration for you.

https://patroni.readthedocs.io/en/latest/replica_bootstrap.html

bkhomuts: It makes sense to set it to True . I will test it with the new value and will post results.

bkhomuts: I have set the value to True, and it works as expected. Let's keep it True.

bkhomuts:
I have tested this change and confirmed it works when the variable `pgbackrest_do_patroni_clean_bootsrap` is set to either true or false.
Log files:
[log_redacted_pgbackrest_do_patroni_clean_bootsrap.log](https://github.com/user-attachments/files/24193269/log_redacted_pgbackrest_do_patroni_clean_bootsrap.log)


[log_redacted_pgbackrest_no_patroni_clean_bootsrap.log](https://github.com/user-attachments/files/24192419/log_redacted_pgbackrest_no_patroni_clean_bootsrap.log)

I have only tested `deploy_pgcluster` on the clean install, when `/var/lib/postgresql` is empty.

If `pgbackrest_do_patroni_clean_bootstrap: true`, then bootstrap is done using patroni, otherwise by the old process (remains default).

For deploy_pgcluster to work, I had to ensure that PostgreSQL data directory is empty, if it wasn't present prior to deploy, otherwise patroni fails to start due to system ID mismatch.

bkhomuts:
I have also tried the PITR recovery (with only tag point_in_time_recovery) now.

It works for both cases, the log files are similar, but in case pgbackrest_do_patroni_clean_bootstrap --- the postgresql data directory is removed prior to recovery.

In my case, the master restored successfully using --delta, and the replica restored from master:
```Dec 16 15:01:00 node2.localdomain patroni[8730]: patroni: INFO:  stderr=pg_rewind: servers diverged at WAL location 2AE/4D4836E8 on timeline 143
Dec 16 15:01:00 node2.localdomain patroni[8730]: pg_rewind: error: could not restore file "0000008F000002AE0000004D" from archive
Dec 16 15:01:00 node2.localdomain patroni[8730]: pg_rewind: error: could not find previous WAL record at 2AE/4D4836E8
Dec 16 15:01:00 node2.localdomain patroni[8730]: patroni: ERROR: Failed to rewind from healthy primary: node1
Dec 16 15:01:00 node2.localdomain patroni[8730]: patroni: WARNING: remove_data_directory_on_rewind_failure is set. removing...
```

automation/roles/common/defaults/main.yml
```diff
@@ -777,9 +777,13 @@ pgbackrest_server_conf:
 # the stanza section will be generated automatically
 
 pgbackrest_archive_command: "pgbackrest --stanza={{ pgbackrest_stanza }} archive-push %p"
-
+pgbackrest_do_patroni_clean_bootsrap: false
```
vitabaks: what is missing here is a brief description of the variable, what it is needed for.
Also, the new variable must be described in the README of the pgbackrest role.

automation/roles/patroni/tasks/main.yml
```diff
@@ -381,6 +381,12 @@
   when: postgresql_wal_dir is defined and postgresql_wal_dir | length > 0
   tags: patroni, custom_wal_dir
 
+- name: Check if data directory exists, before creating it
+  ansible.builtin.stat:
+    path: "{{ postgresql_data_dir }}"
+  register: stat_postgresql_data_dir
+  tags: patroni, point_in_time_recovery
```
vitabaks: It seems to me that this task is optional.

automation/roles/patroni/tasks/main.yml
```diff
@@ -464,17 +470,6 @@
             group: postgres
             mode: "0755"
 
-        - name: Prepare PostgreSQL | make sure the data directory "{{ postgresql_data_dir | default('') }}" is empty
```
vitabaks, line 467: See the comment below

automation/roles/patroni/tasks/main.yml
```diff
@@ -502,10 +497,24 @@
             - absent
             - directory
           when: postgresql_wal_dir | default('') | length > 0
-      when: patroni_cluster_bootstrap_method != "pgbackrest"
+      when: patroni_cluster_bootstrap_method != "pgbackrest" or pgbackrest_do_patroni_clean_bootsrap | default(false)
   when: not postgresql_exists | default(false) | bool or patroni_cluster_bootstrap_method != "initdb"
   tags: patroni, point_in_time_recovery
 
+- name: Prepare PostgreSQL | make sure the data directory "{{ postgresql_data_dir | default('') }}" is empty
```
vitabaks: I'm not sure if there's a need to move this task from the block above.

vitabaks:
@bkhomuts I renamed pgbackrest_do_patroni_clean_bootstrap to `pgbackrest_patroni_cluster_clean_bootstrap`

to make it closer to the already existing style of pgbackrest_patroni_cluster_xx variables.

automation/roles/patroni/tasks/main.yml
bkhomuts: That task is needed to determine if the directory was empty prior to pg_createcluster.

If the data directory contained files (empty DB), the cluster won't bootstrap from Patroni --- an error "system ID mismatch" will be returned. To avoid such error, it is necessary to clean the just created data directory by pg_createcluster.

automation/roles/patroni/tasks/main.yml
bkhomuts: This is related to my previous comment https://github.com/vitabaks/autobase/pull/1384#discussion_r2626882217. To correctly bootstrap the cluster from patroni using pgbackrest, it is necessary to ensure that the directory is empty, otherwise, there will be system ID mismatch error.

bkhomuts:
Thanks for a review, @vitabaks .

I left brief answers. If needed, I can explain the issue in more detail, and attach logs.

automation/roles/patroni/tasks/main.yml
vitabaks: this directory cleanup task would have been executed within the block above, thanks to condition `when: patroni_cluster_bootstrap_method != "pgbackrest" or pgbackrest_patroni_cluster_clean_bootstrap | default(false)`. 

in addition, the current change will break PITR for WAL-G because the directory will not be cleared.

automation/roles/patroni/tasks/main.yml
vitabaks: we already have a task to clean up the directory if the pgbackrest_patroni_cluster_clean_bootstrap variable is set, so there shouldn't be any problems even without this check.

bkhomuts: Thanks for pointing out, I will check it again tomorrow. Maybe this change isn't needed.

bkhomuts: I rerun the playbooks again, and you were right, having only `pgbackrest_patroni_cluster_clean_bootstrap: true` is sufficient.

I have removed this extra check, which remained in the process of developing and testing this change. Again, thanks for pointing this out.

automation/roles/patroni/tasks/main.yml
bkhomuts: I have returned the task back to its block.

automation/roles/patroni/tasks/main.yml
bkhomuts: I have returned this task back.

Issue #1392, opened 2025-12-17
[Bug] Upgrade fails with wal_level: logical


bkhomuts:
### Bug description

When wal_level is set to logical, upgrade fails. This is because the new PostgreSQL DB is created with wal_level: replica by default. As a result, pg_upgradecluster returns an error due to wal_level mismatch.

### Expected behavior

pg_upgradecluster succeeds. This requires wal_level on the new cluster to be set to the same value as the old.

### Installation method

Console (UI)

### System info

Debian 13
Autobase from Git

### Additional info

Add
```-c wal_level={{ current_wal_level }}```
to pg_createcluster and initdb when upgrading.

vitabaks:
Just adjust the parameter values on both servers to the same value.

bkhomuts:
I have used temporary workaround by adding  `-c wal_level=logical` for the task
Initialize new PostgreSQL data directory with default config files, and the upgrade succeeded.

But I think this needs to be accounted for in automation.

vitabaks:
Ok, feel free to suggest a PR.

PR #1394, opened 2025-12-17
Support pg_upgrade with wal_level: logical

bkhomuts:
This change ensures that the new cluster is created with the same wal_level, as the old one.

The change allows cluster upgrade if wal_level: logical.

Resolves: #1392


bkhomuts:
The last CI run issue is not related to the change, but problem on servers.

Issue #1390, opened 2025-12-16
[Bug] pg_upgrade fails when non-standard install user with password is used


bkhomuts:
### Bug description

It is not possible to run vitabaks.autobase.pg_upgrade if the install user is not postgres or if the install user is protected with password.

In some cases, it is necessary to specify password.

### Expected behavior

Successful upgrade, even if the postgres (or non-standard install user requires password).

### Installation method

Command line

### System info

Debian 13
PostgreSQL upgrade from 17 to 18

### Additional info

I have a working solution at:
https://github.com/bkhomuts/autobase/commit/779259d32a2ce0ee06b9354f5420d813f276a674

Which I will submit as a PR, once I finalize it.

Issue #1387, opened 2025-12-14
[Bug] pg_upgrade hangs indefinitely - pgbouncer issue


violuke:
### Bug description

When trying to upgrade from pg 17 to 18 (or in my case, run the pre-upgrade checks as we've not managed to pass those yet), it hangs at:

```
TASK [vitabaks.autobase.upgrade : [Pre-Check] Test PgBouncer access via unix socket]
```

I *think* this is due to `PGPASSWORD` not being set in this task like it is in many others:
```
  environment:
    PGPASSWORD: "{{ patroni_superuser_password }}"
```

Again, I might be wrong, but I think the same will need to be done in [the actual PAUSE command](https://github.com/vitabaks/autobase/blob/1ed7cb223b4465024b595d602699d28a8d90aff8/automation/roles/upgrade/tasks/pgbouncer_pause.yml#L33), and maybe other places.

Any help would be much appreciated. Thank you.

### Expected behavior

Not to hang

### Steps to reproduce

```
ansible-playbook -i inventory.yml vitabaks.autobase.pg_upgrade -e "target=postgres pg_old_version=17 pg_new_version=18" --tags "pre-checks,upgrade-check"
```

### Installation method

Command line

### System info

- Ubuntu 24.04.3 LTS
- Autobase 2.5.2

vitabaks:
Hi! Please attach pgbouncer_pool_pause log in /tmp directory.

violuke:
I don't have the logs, I'm sorry. I might be wrong but I don't think there would have been a log to there for the `[Pre-Check] Test PgBouncer access via unix socket]` task? That would be from the actual pause right?

I was able to bypass this issue by running my own playbook instead that created a `.pgpass` file for the PgBouncer unix stock connection to get its password from.

```yaml
- name: Prepare .pgpass for upgrade
  hosts: all
  tags: [pre-checks, upgrade-check, upgrade, always]
  vars:
    _pgbouncer_port: "{{ pgbouncer_listen_port | default(6432) }}"
  tasks:
    - name: Check if .pgpass exists
      stat:
        path: ~/.pgpass
      register: pgpass_stat
      become: yes
      become_user: postgres

    - name: Read existing .pgpass content
      slurp:
        src: ~/.pgpass
      register: pgpass_content
      become: yes
      become_user: postgres
      when: pgpass_stat.stat.exists

    - name: Check if pgbouncer line already exists
      set_fact:
[...]

PR #1395, opened 2025-12-17
Support upgrade_pgcluster with custom install user

bkhomuts:
Support upgrade of the cluster where install_user is not postgres and where password is required.

Introduce pg_install_user_password variable --- in case pg_install_user is not patroni superuser.

Do not pass GRANTs to new PostgreSQL version during schema check, this prevents errors related to permissions
(not related to schema).

Resolves: #1390

### DIFF
--- a/automation/roles/upgrade/tasks/initdb.yml
+++ b/automation/roles/upgrade/tasks/initdb.yml
@@ -82,7 +82,7 @@
 # for Debian based use pg_createcluster, if the PostgreSQL configuration is not located in the data directory.
 # Note: Patroni failure is possible if the default postgresql config files are missing in the /etc/postgresql/...
 - name: Initialize new PostgreSQL data directory with default config files
-  ansible.builtin.shell: >
+  ansible.builtin.command: >
     /usr/bin/pg_createcluster {{ pg_new_version }} {{ postgresql_cluster_name }}
     --datadir={{ pg_new_datadir }}
     --encoding={{ (hostvars[groups['primary'][0]].pg_settings.stdout | from_json).server_encoding }}
@@ -102,8 +102,6 @@
     {% if (hostvars[groups['primary'][0]].pg_settings.stdout | from_json).data_checksums == 'on' %}
     --data-checksums
     {% endif %}
-  args:
-    executable: /bin/bash
   when:
     - ansible_os_family == "Debian"
     - pg_new_confdir != pg_new_datadir


automation/roles/upgrade/tasks/initdb.yml
```diff
@@ -71,12 +71,19 @@
...
+    content: "{{ pg_install_user_password }}"
+    mode: "0600"
+  no_log: true
+  when: pg_install_user_password is defined
+
 # for Debian based use pg_createcluster, if the PostgreSQL configuration is not located in the data directory.
 # Note: Patroni failure is possible if the default postgresql config files are missing in the /etc/postgresql/...
 - name: Initialize new PostgreSQL data directory with default config files
-  ansible.builtin.command: >
+  ansible.builtin.shell: >
```
vitabaks: Isn't the `command` module suitable here?

automation/roles/upgrade/tasks/upgrade_primary.yml
```diff
@@ -17,6 +17,8 @@
     --link
   args:
     chdir: "/tmp"
+  environment:
+    PGPASSWORD: "{{ pg_install_user_password | default(patroni_superuser_password) }}"
```
vitabaks, line 21: > Perhaps we should add a password file check instead of using the PGPASSWORD variable.

See https://github.com/vitabaks/autobase/issues/1387

automation/roles/upgrade/tasks/initdb.yml
```diff
@@ -71,12 +71,19 @@
   when:
     - inventory_hostname in groups['primary']
 
+- name: Create temporary password file for install user
```
vitabaks, line 74: I'm not sure if we need a custom password file, instead we should make sure that there is an entry in the existing .pgpass file (using the [pgpass](https://github.com/vitabaks/autobase/tree/master/automation/roles/pgpass) role).

This may be part of [pre_checks.yml](https://github.com/vitabaks/autobase/blob/master/automation/roles/upgrade/tasks/pre_checks.yml)

bkhomuts: @vitabaks , 

`.pgpass` is used when connecting to the cluster, but not when doing `pg_createcluster`. If password isn't passed, there will be problem logging into the new cluster (when a custom install user is used --- so it is not possible to use peer authentication).

automation/roles/upgrade/tasks/initdb.yml
bkhomuts: That's right, command is sufficient here.

Initially, I tried to avoid creating a temporary file using shell to pass string as a file, but it didn't work with pg_createcluster. As the redirect is no longer used, shell is no not needed.

automation/roles/upgrade/tasks/upgrade_primary.yml
bkhomuts: That would be technically correct solution.

bkhomuts: I'd prefer to leave it as it is for this MR, as it is consistent with how password is passed in other places, and have a separate MR for the problem described in #1387 .

Issue #1420, opened 2026-01-07
[Bug] When deploying PostgreSQL 18 , data checksums can't be disabled


bkhomuts:
### Bug description

When deploying PostgreSQL 18 with `postgresql_data_checksums: false`, the data checksums are enabled.

### Expected behavior

Data checksums are disabled when postgresql_data_checksums: false and during upgrade from 17 -> 18.

### Steps to reproduce

1. Deploy cluster with `postgresql_version: 18` and `postgresql_data_checksums: false`
2. SELECT name,setting FROM pg_settings WHERE name = 'data_checksums';

### Installation method

Command line

### System info

Debian 13
Autobase from Git.

### Additional info

Current logic assumes that data checksums are disabled by default. Since PostgreSQL 18, data checksums are enabled, unless explicitly disabled. Therefore, Autobase needs to be updated to specify `--no-data-checksums` for PostgreSQL >= 18, when the postgresql_data_checksums: false, and during upgrade.

PR #1421, opened 2026-01-07
Update data-checksums logic for PostgreSQL 18

bkhomuts:
With data-checksums are now  enabled by default, it is necessary to update the logic for initdb and upgrade if PostgreSQL >= 18.

Resolves: #1420


vitabaks:
Please add `int` filter for `postgresql_version` variable,

bkhomuts:
Added.

Issue #1422, opened 2026-01-08
Allow disabling trust authentication for local connections for the replication user


bkhomuts:
### Problem

At the moment pg_hba.conf is created with trust authentication for the replication user (replicator). This is not required for all configurations, as Patroni connects using username/password.

### Description

There should be a configuration option to disable trust authentication for the replication user of the cluster, as sometimes, password authentication is preferred over 'trust'.

### Importance

must have

### Proposed implementation

Update pg_hba.conf.j2

PR #1423, opened 2026-01-08
Require authentication for local replication connections

bkhomuts:
As patroni logins with username/password, trust authentication is not required.

Resolves: #1422


vitabaks:
As far as I remember, there were problems if we didn't add trust for local, now I can't remember for sure, maybe the cluster not working during [cloning](https://autobase.tech/docs/management/restore#cloning) when a different replication user password is specified.

This may be relevant for older versions of Patroni.

vitabaks:
The `postgresql_pg_hba` variable also includes several trust rules for local connections. I assumed that access to the database server itself is already restricted to authorized users. If an attacker manages to get onto the server, this change is unlikely to provide much additional protection anyway.

bkhomuts:
@vitabaks , thanks for providing additional context.

I do think applying the principle of least privilege is important for the replicator. Without this change, any unprivileged user can log-into the replication user from the local machine. But requiring password will protect PostgreSQL data if another service on the machine is breached.

The compatibility with older versions is valid concern, I'll then make this option configurable.

On my systems, `postgresql_pg_hba` doesn't have `trust` for users. It is possible to run hardened configuration --- better to provide choice to the end user.

PR #1433, opened 2026-01-21
Fix the postgres slots layout in the patroni configuration

zhetigen-abdisattarov:
Fixes: https://github.com/vitabaks/autobase/issues/1432

### DIFF
--- a/automation/roles/patroni/README.md
+++ b/automation/roles/patroni/README.md
@@ -121,16 +121,19 @@ Requirements:
 1. the cluster name for standby cluster must be unique (see 'patroni_cluster_name' variable)
 2. the IP addresses (or network) of the standby cluster servers must be added to the pg_hba.conf of the Main Cluster (see 'postgresql_pg_hba' variable).
 
-### Slots Configuration
+### Permanent Replication Slots Configuration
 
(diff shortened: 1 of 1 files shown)


automation/roles/common/defaults/main.yml
```diff
@@ -553,10 +553,10 @@ patroni_standby_cluster:
 #  restore_command: ""  # command to restore WAL records from the remote master to standby leader (optional)
 #  recovery_min_apply_delay: ""  # how long to wait before actually apply WAL records on a standby leader (optional)
 
-# Permanent replication slots.
-# These slots will be preserved during switchover/failover.
+# Permanent replication slots (stored in DCS at the 'slots' level).
+# These slots will be preserved and automatically recreated during switchover/failover.
 # https://patroni.readthedocs.io/en/latest/dynamic_configuration.html
-patroni_slots: []
+patroni_permanent_slots: []
```
vitabaks, line 559: I suggest not renaming the variable, so we can achieve the result with fewer changes.


The fact that these are permanent slots is indicated in the comment.

vitabaks:
https://github.com/vitabaks/autobase/pull/1445

Issue #1417, opened 2026-01-06
[Feat] Allow installing etcd from the package repositories (apt or yum)


bkhomuts:
### Problem

etcd is available in Debian as a package, but there is no option to install it from the repository.

### Description

Allow installing etcd from the Debian repository.

### Importance

nice to have

PR #1434, opened 2026-01-21
Allow installing etcd from system repositories

bkhomuts:
When etcd_installation_method: {{ pkg_type }}, then etcd is installed from system (apt or yum) repositories.

Etcd is installed from GitHub repository by default.

This change also enables the role `add_repository` on `etcd_cluster` members, enabling by default PGDG (and PGDG extras, EPEL and CRB on RHEL-based systems).

Resolves: #1417

### DIFF
--- a/automation/roles/common/defaults/main.yml
+++ b/automation/roles/common/defaults/main.yml
@@ -124,7 +124,7 @@ dcs_exists: false # or 'true' if you do not want Autobase to deploy and manage e
 etcd_version: 3.5.25
 # Note: when etcd_installation_method: "rpm", the etcd package is installed from pgdg-rhel-extras,
 # unless install_postgresql_repo or install_postgresql_repo_extras are set to false.
-etcd_installation_method: "{{ pkg_type }}" # binary -> install from repo such as https://github.com, "{{ pkg_type }}" -> install from apt or yum repository
+etcd_installation_method: "binary" # binary -> install from repo such as https://github.com, "{{ pkg_type }}" -> install from apt or yum repository
 etcd_bin_dir: "{{ '/usr/bin' if etcd_installation_method == pkg_type else '/usr/local/bin' }}"
 etcd_conf_dir: "/etc/etcd"
 etcd_data_dir: "/var/lib/etcd"


vitabaks:
@bkhomuts Thanks for PR!

It looks like there is no package for RHEL-based.

>   fatal: [pgnode01]: FAILED! => {"attempts": 3, "changed": false, "failures": ["No package etcd available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}

Test: https://github.com/vitabaks/autobase/actions/runs/21217989420/job/61044550368

bkhomuts:
@vitabaks , yes, it is not available in the standard RHEL repositories. Only available in Fedora.


I have tested this change only in Debian.

vitabaks:
In this case, we need a [pre-check](https://github.com/vitabaks/autobase/tree/master/automation/roles/pre_checks/tasks) for etcd so that it stops the playbook if etcd_installation_method is selected with a value equal yum.

bkhomuts:
For RHEL, the package is available in PGDG extras:
```sudo dnf config-manager --enable pgdg-rhel10-extras```

**Comment by vitabaks** [comment]
then it's enough to add this task for yum:
```yaml
    - name: Enable PostgreSQL extras repository
      ansible.builtin.command: dnf config-manager --enable pgdg-rhel{{ ansible_distribution_major_version }}-extras
      register: pgdg_extras_result
      failed_when:
        - pgdg_extras_result.rc != 0
        - "'No matching repo' not in pgdg_extras_result.stderr"
```

https://github.com/vitabaks/autobase/blob/master/automation/roles/add_repository/tasks/main.yml#L132

automation/roles/etcd/tasks/main.yml
```diff
@@ -121,6 +121,33 @@
...
+- block:
+    - name: Enable PGDG extras repository for etcd
+      # etcd needs PGDG extras repository on RHEL, as etcd isn't available from standard repositories.
+      ansible.builtin.import_role:
+        name: vitabaks.autobase.add_repository
+        allow_duplicates: false
+      when:
+        - etcd_installation_method == 'rpm'
+        - install_postgresql_repo_extras | bool
+        - install_postgresql_repo | bool
```
vitabaks:
```suggestion
    - name: Enable PGDG extras repository for etcd
      ansible.builtin.command: dnf config-manager --enable pgdg-rhel{{ ansible_distribution_major_version }}-extras
      register: pgdg_extras_result
      failed_when:
        - pgdg_extras_result.rc != 0
        - "'No matching repo' not in pgdg_extras_result.stderr"
      when:
        - etcd_installation_method == 'rpm'
        - install_postgresql_repo_extras | bool
```

automation/roles/common/defaults/main.yml
```diff
@@ -122,6 +122,10 @@ dcs_exists: false # or 'true' if you do not want Autobase to deploy and manage e
 # if dcs_type: "etcd" and dcs_exists: false
 # renovate: datasource=github-releases depName=etcd-io/etcd extractVersion=^v(?<version>.*)$
 etcd_version: 3.5.25
+# NOTE: when etcd_installation_method: "{{ pkg_type }}" or "rpm" for RHEL-based OS, ensure that etcd is available from repos.
+# This can be achieved by enabling PGDG extras repo by setting variables to true: install_postgresql_repo and install_postgresql_repo_extras
```
vitabaks:
```suggestion
# Note: when etcd_installation_method: "rpm", the etcd package is installed from pgdg-rhel-extras.
```

bkhomuts:
> > fatal: [pgnode01]: FAILED! => {"attempts": 3, "changed": false, "failures": ["No package etcd available."], "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
> 
> Test: https://github.com/vitabaks/autobase/actions/runs/21217989420/job/61044550368

@vitabaks , thanks for a quick review.

I thought more about this, I would rather solve it  by including `add_repository` role into the etcd_cluster, similarly to how is it done in pgbackrest_install_from_pgdg_repo . This way we avoid contradiction with `install_postgresql_repo_extras`. This will cover a use case when a user would like to install etcd from its own repo, but doesn't want to enable the PGDG extras repo.

bkhomuts:
I'm now testing this change in my local environment, and will update the PR once done.

vitabaks:
> I would rather solve it by including add_repository role into the etcd_cluster

Adding the add_repository role is a bit superfluous here, since repositories for the Postgres cluster may be declared there, rather than for the ETCD cluster, which may be on dedicated servers.

bkhomuts:
> Adding the add_repository role is a bit superfluous here, since repositories for the Postgres cluster may be declared there, rather than for the ETCD cluster, which may be on dedicated servers.

In this case, then adding an additional check for `install_postgresql_repo_extras` may be sufficient. But I need to verify if `sudo dnf config-manager --enable pgdg-rhel10-extras` will be sufficient. This is because PGDG RPM has to be installed first.

bkhomuts:
So I tested this change both with import_role and with individual tasks. With import_role (commit 6bbd3a28ef264686a42a4a99acb3a0aed8983582) the solution is cleaner, but it also enables CRB and installs EPEL.

With individual tasks, no additional repos are enabled, but the downside is that the user can't use yum_repos to get etcd from there.

I'd personally prefer the import_role solution, and if extra repositories are unneeded at etcd servers, just set the corresponding variables.

vitabaks:
Thanks for testing!

Alternatively, we can add the add_repository role to the etcd_cluster.yml playbook.

There are different ways to do this, the main thing is that it works.

bkhomuts:
> Alternatively, we can add the add_repository role to the etcd_cluster.yml playbook.

Done. I ran the test deployment both on AlmaLinux 10 and Debian 13. Worked fine.

vitabaks:
> I ran the test deployment both on AlmaLinux 10 and Debian 13. Worked fine.

hmm, the Molecule tests failed.

https://github.com/vitabaks/autobase/actions/runs/21400821137/job/61611071037?pr=1434

bkhomuts:
> > I ran the test deployment both on AlmaLinux 10 and Debian 13. Worked fine.
> 
> hmm, the Molecule tests failed.
> 
> https://github.com/vitabaks/autobase/actions/runs/21400821137/job/61611071037?pr=1434

This could be due to TLS. I remember seeing a similar issue, so I set in my configuration `etcd_tls_enable: false`. Will try with default configuration.

bkhomuts:
The problem was in `/etc/etcd` directory permissions being reset after installing etcd from Yum. Solved by setting the permissions to the expected value after etcd installation from yum repository on RHEL.

The issue is that permissions will reset on etcd upgrade. In my opinion, that is a problem of the RPM package itself --- its spec file needs to be adjusted.

PR #1459, opened 2026-02-21
Preserve per-node patroni_tags in start_traffic.yml during updates

jimnyzuk:
`start_traffic.yml` unconditionally reset `noloadbalance`, `nosync`, and `nofailover` to `false` after updates, overwriting intentional per-node settings from `patroni_tags` inventory variable.


vitabaks:
Ok, thanks

PR #1507, opened 2026-04-09
fix: pgbackrest cron jobs fail on dedicated backup server

MetalBlueberry:
## Summary

- Scheduled pgbackrest backups never run when cron jobs are installed on a dedicated backup server (repo_host)
- The `pg_is_in_recovery()` psql guard in the cron job fails silently because there is no local PostgreSQL instance on the backup server
- Added `pgbackrest_cron_jobs_repo` variable without the guard, used when cron is installed on the dedicated pgbackrest server

## Details

The existing `pgbackrest_cron_jobs` wraps the backup command in:
```bash
if [ $(psql -tAXc 'select pg_is_in_recovery()') = 'f' ]; then pgbackrest ... backup; fi
```

This guard makes sense on database servers (prevents replicas from triggering backups), but on a dedicated backup server there is no local PostgreSQL — `psql` fails, the condition is never true, and backups never run.

The only backup that exists after initial deployment is the one triggered by the playbook during `stanza-create`. All subsequent scheduled backups are silently skipped.

## Test plan

- [x] Deploy a cluster with a dedicated pgbackrest backup server
- [x] Verify `/etc/cron.d/pgbackrest-<cluster>` on the backup server does NOT contain the `psql` guard
- [x] Wait for cron to fire and confirm a new backup appears in `pgbackrest info`

### DIFF
--- a/automation/roles/common/defaults/main.yml
+++ b/automation/roles/common/defaults/main.yml
@@ -831,6 +831,20 @@ pgbackrest_patroni_cluster_clean_bootstrap: false  # false: use delta restore fo
 
 # By default, the cron jobs is created on the database server.
 # If 'repo_host' is defined, the cron jobs will be created on the pgbackrest server.
+#
+# On database servers, the pg_is_in_recovery() guard prevents replicas from triggering backups.
+# On a dedicated backup server (repo_host), the guard is omitted because there is no local
+# PostgreSQL instance — psql would fail silently and backups would never run.
+pgbackrest_full_backup_command: >-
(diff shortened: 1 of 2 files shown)


vitabaks:
Alternatively, we could define variables like `pgbackrest_full_backup_command` and populate them dynamically depending on whether pgbackrest_repo_host is set.

vitabaks:
@MetalBlueberry what do you think about my comment above?

MetalBlueberry:
Something like the changes in the latest commit? 
I do not have strong opinions here. Both work

vitabaks:
Yes, I think this is better because it allows you to override the backup command without overriding the entire cron variable.

PR #1553, opened 2026-05-21
fix: resolve nested Jinja2 templating error in postgresql_index_maintenance

fasiihullah:
The 'enabled' key inside the postgresql_index_maintenance dict was defined as a Jinja2 template referencing postgresql_index_maintenance_enabled. When Ansible evaluated postgresql_index_maintenance.enabled in a 'when' condition, it received a raw unresolved template string and failed with:

  "The conditional check 'postgresql_index_maintenance.enabled | default(false) | bool'
   failed. Error was: Jinja2 error: nested Jinja2 templating"

Fix:
- Extract postgresql_index_maintenance_enabled as a top-level variable in common/defaults/main.yml (default: true) so it is cleanly overridable in group_vars without needing to redefine the entire dict
- Replace all 'when: postgresql_index_maintenance.enabled | default(false) | bool' conditionals with 'when: postgresql_index_maintenance_enabled | bool'

Users can now disable index maintenance with a single line in group_vars:
  postgresql_index_maintenance_enabled: false

### DIFF
--- a/automation/playbooks/add_node.yml
+++ b/automation/playbooks/add_node.yml
@@ -341,7 +341,7 @@
         (cluster_vip is defined and cluster_vip | length > 0)
 
     - role: vitabaks.autobase.postgresql_index_maintenance
-      when: postgresql_index_maintenance_enabled | bool
+      when: postgresql_index_maintenance_enabled | default(false) | bool
 
     - role: vitabaks.autobase.netdata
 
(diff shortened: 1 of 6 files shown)


vitabaks:
@fasiihullah Thanks for the PR. 

Although I do not recommend disabling index maintenance, it is important for database performance.

PR #1554, opened 2026-05-22
feat(console): support Docker Secrets via _FILE convention

pen-pal:
Related to #1480

Adopt the convention popularized by the official postgres image so the console can consume sensitive values (auth token, database password) from files mounted under /run/secrets instead of plain-text env vars in .env.

For any FOO, if FOO_FILE is set the file's contents are loaded as FOO. A single trailing newline is trimmed. Setting both FOO and FOO_FILE non-empty is an error.

- Go API: resolveFileSecrets sweeps PG_CONSOLE_*_FILE env vars before envconfig.Process runs; covers AUTHORIZATION_TOKEN, DB_PASSWORD, and ENCRYPTIONKEY uniformly.
- UI: POSIX-sh file_env helper in env.sh applied to PG_CONSOLE_AUTHORIZATION_TOKEN so the value the script injects into the built JS comes from the secret file.
- DB: bash file_env helper in pg_start.sh applied to POSTGRES_PASSWORD so the ALTER USER statement sources its value from the secret file.
- docker-compose.secrets.yml: opt-in override demonstrating Docker Secrets; also overrides the API healthcheck to read the bearer token from the mounted secret at exec time, so the cleartext never appears in the container's environment or `docker inspect` output.
- Config.Redacted(): the startup config dump in main.go was writing secret values to logs in cleartext, which defeats the purpose of using Docker Secrets. The logged copy now masks Token/Password/EncryptionKey as "[REDACTED]" (non-empty) or "[unset]" (empty).

End-to-end verified by building all three console images locally, bringing the stack up with the secrets override, and confirming: API resolves both secrets, DB password is enforced over TCP (scram-sha-256), bearer auth works with the file-backed token, conflict and missing-file paths fail cleanly, and the cleartext values no longer appear in logs.

Note: console/db/Dockerfile bakes ENV POSTGRES_PASSWORD=postgres-pass, so adopters must explicitly set POSTGRES_PASSWORD: "" alongside _FILE (the override file does this) to avoid hitting the conflict guard. Out of scope to remove that Dockerfile default here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ __pycache__/
 molecule/**/tmp
 .ansible
 .ansible-lint-env
+# Docker Secrets directory created by docker-compose.secrets.yml
+console/secrets/
--- a/console/README.md
+++ b/console/README.md
@@ -86,3 +86,27 @@ The Console stack consists of the following core components:
 - The Caddy network is created automatically by Docker Compose
 - All services are configured to restart automatically unless stopped manually.
 - Additional [environment variables](https://github.com/autobase-tech/autobase/tree/main/console/service#configuration) can be configured based on your project needs.
+
+## Using Docker Secrets
+
+Sensitive values (auth token, database password) can be supplied via files
+mounted under `/run/secrets` instead of plain-text env vars in `.env`. Every
+console image honors the `_FILE` convention used by the official postgres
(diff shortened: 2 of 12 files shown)


vitabaks:
Hi! Thanks for the contribution.

PR #1557, opened 2026-05-23
fix owner of /var/log/pgbouncer for pgbouncer.service

vreitech:
Fixes problem with owner on `/var/log/pgbouncer` after pgbouncer package's update on RH-based distros.
Every time pgbouncer package updates on AlmaLinux 8/9, the owner of `/var/log/pgbouncer` directory changes to `pgbouncer`, and it prevents running `pgbouncer.service` because it needs the directory have to belongs to `postgres` user.

### DIFF
--- a/automation/roles/pgbouncer/templates/pgbouncer.service.j2
+++ b/automation/roles/pgbouncer/templates/pgbouncer.service.j2
@@ -16,7 +16,7 @@ ExecStart=/usr/sbin/pgbouncer -d {{ pgbouncer_conf_dir }}/pgbouncer{{ '-%d' % (i
 {% endif %}
 {% if ansible_os_family == "RedHat" %}
 PermissionsStartOnly=true
-ExecStartPre=/usr/bin/chown -R postgres: /var/log/pgbouncer
+ExecStartPre=/usr/bin/chown -R postgres: {{ pgbouncer_log_dir }}
 ExecStart=/usr/bin/pgbouncer -d {{ pgbouncer_conf_dir }}/pgbouncer{{ '-%d' % (idx + 1) if idx > 0 else '' }}.ini
 {% endif %}
 ExecReload=/bin/kill -SIGHUP $MAINPID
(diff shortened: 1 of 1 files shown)


vreitech:
[Probably] related issue: https://github.com/autobase-tech/autobase/issues/636

Yes, it's looks mostly like workaround. Probably better solution would be to start pgbouncer using `pgbouncer` OS user, and probably change a bit `pg_hba.conf` or `pgbouncer.ini` (not sure it needed btw). I just propose easiest one solution i've found, feel free to discuss/propose/etc if interested.

automation/roles/pgbouncer/templates/pgbouncer.service.j2
```diff
@@ -15,6 +15,8 @@ RuntimeDirectoryMode=0755
 ExecStart=/usr/sbin/pgbouncer -d {{ pgbouncer_conf_dir }}/pgbouncer{{ '-%d' % (idx + 1) if idx > 0 else '' }}.ini
 {% endif %}
 {% if ansible_os_family == "RedHat" %}
+PermissionsStartOnly=true
+ExecStartPre=/usr/bin/chown -R postgres: /var/log/pgbouncer
```
vitabaks:
```suggestion
ExecStartPre=/usr/bin/chown -R postgres: {{ pgbouncer_log_dir }}
```

vitabaks:
Ok, thanks @vreitech !

automation/roles/pgbouncer/templates/pgbouncer.service.j2
```diff
@@ -15,6 +15,8 @@ RuntimeDirectoryMode=0755
 ExecStart=/usr/sbin/pgbouncer -d {{ pgbouncer_conf_dir }}/pgbouncer{{ '-%d' % (idx + 1) if idx > 0 else '' }}.ini
 {% endif %}
 {% if ansible_os_family == "RedHat" %}
+PermissionsStartOnly=true
+ExecStartPre=/usr/bin/chown -R postgres: {{ pgbouncer_log_dir }}
```
vreitech, line 19: Just as remindment: `postgres:` in terms of `chown` means "to change owner and group to 'postgres'".

Issue #1516, opened 2026-04-20
[Feat]  Haproxy behaviour tuning


a-b-v:
### Problem

Long queries are abruptly terminated

### Description

The HAProxy template includes:

`default-server inter 3s fastinter 1s fall 3 rise 4 on-marked-down shutdown-sessions{% if patroni_restapi_protocol == 'https' %} check-ssl verify none{% endif %}`

This ensures outdated replicas are not used, but it also causes long-running queries to be terminated when a replica is marked down.

### Importance

nice to have

### Proposed implementation

Add an option to make this behavior configurable (e.g., disable or customize shutdown-sessions), so users can avoid interrupting long-running queries when needed.

vitabaks:
@a-b-v How long are your queries? Do they exceed [haproxy_timeout](https://github.com/autobase-tech/autobase/blob/2.7.0/automation/roles/common/defaults/main.yml#L51)?

a-b-v:
No, they’re under 30 minutes, but replica lag can exceed the limit during that period

vitabaks:
See `patroni_maximum_lag_on_replica` variable.

a-b-v:
I’ve already adjusted this parameter, but I still think it would be useful to have the ability to disable this behavior.

a-b-v:
So that new requests are executed on an up-to-date dataset, while allowing older long-running ones to finish.

vitabaks:
Feel free to suggest PR

vitabaks:
Done https://github.com/autobase-tech/autobase/pull/1562

PR #1562, opened 2026-05-29
feat/haproxy shutdown sessions toggle

pen-pal:
## Summary

Make HAProxy's per-server health-check / failover behavior configurable through a new `haproxy_default_server` variable. The main knob is `on_marked_down`: set it to `""` to drop `on-marked-down shutdown-sessions`, so active sessions survive when a replica is marked down instead of being killed mid-query.

Associates #1516.

## Motivation

The `default-server` line was hardcoded, identically, across both HAProxy templates:

```
default-server inter 3s fastinter 1s fall 3 rise 4 on-marked-down shutdown-sessions
```

A backend is marked **down** as soon as its health check fails a few times in a row — which can happen for harmless reasons, such as a replica briefly falling behind on replication lag or a momentary network blip. The `on-marked-down shutdown-sessions` part tells HAProxy that, the instant this happens, it should immediately drop every open connection to that backend.

The side effect: any query still running on that replica is cut off mid-execution. A long analytics query can die seconds in, just because the replica blipped once. And there was no way to turn this off without editing the role's templates.

## Changes

- **`common/defaults/main.yml`** — add the `haproxy_default_server` dict (defaults below, reproducing the previous hardcoded values exactly).
- **`haproxy/templates/haproxy.cfg.j2`** — replace the 8 hardcoded `default-server` lines with a single `default_server()` Jinja macro. `on-marked-down` is rendered only when `on_marked_down` is non-empty; `check-ssl verify none` is still emitted for HTTPS.
- **`confd/templates/haproxy.tmpl.j2`** — apply the same macro to the confd-managed copy, keeping the dynamic config in sync with the static one.
- **`haproxy/README.md`** — document the new variables.

### Defaults

| Key | Default | Meaning |
|-----|---------|---------|
| `inter` | `3s` | Interval between health checks |
| `fastinter` | `1s` | Check interval while a server is in transition |
| `fall` | `3` | Failed checks before a server is marked down |
| `rise_master` | `4` | Successful checks before the master backend is marked up |
| `rise_replica` | `2` | Successful checks before a replica backend is marked up |
| `on_marked_down` | `shutdown-sessions` | Action on markdown; `""` keeps sessions alive |

### Backward compatibility

With defaults unchanged, the rendered config is byte-identical to before — nothing changes on upgrade.

To preserve long-running queries:

```yaml
haproxy_default_server:
  on_marked_down: ""   # keep active sessions alive on replica markdown
```

The timing knobs (`inter`, `fastinter`, `fall`, `rise_*`) are now tunable as well, for faster or more conservative failover.

### 1. Rendered config is byte-identical to current behavior

Jinja rendered with Ansible's `trim_blocks=True`, compared against `main` across 6 contexts for both templates:

```
PASS [haproxy.cfg.j2/default]          byte-identical to HEAD
PASS [haproxy.cfg.j2/https]            byte-identical to HEAD
PASS [haproxy.cfg.j2/vip]              byte-identical to HEAD
PASS [haproxy.cfg.j2/pgbouncer]        byte-identical to HEAD
PASS [haproxy.cfg.j2/pgbouncer+direct] byte-identical to HEAD
PASS [haproxy.cfg.j2/balancer_tags]    byte-identical to HEAD
PASS [confd/*]                         byte-identical to HEAD  (all 6 contexts)
PASS on_marked_down='' -> no on-marked-down token
PASS custom inter/fall/rise flow through
```

### 2. Generated config validates against real HAProxy 3.0 (`haproxy -c`)

```
default cfg                                  -> rc=0
keep    cfg  (on_marked_down="")             -> rc=0
custom  cfg  (inter 5s fall 5 rise 9/7)      -> rc=0
confd   cfg                                  -> rc=0
negative control (on-marked-down BOGUSVALUE) -> rc=1   # ALERT raised; harness genuinely validates
```

### 3. Live A/B markdown test

Real HAProxy 3.0 + a backend on a docker network. A long-lived TCP session is held through the load balancer, then the backend health check is flipped to failing. Stats CSV confirms `pg1` transitions `UP -> DOWN` in both runs.

| Variant | `default-server` directive | Result after markdown |
|---------|----------------------------|------------------------|
| A (default) | `... on-marked-down shutdown-sessions` | `SESSION_CLOSED after 3.3s` |
| B (`on_marked_down=""`) | `... rise 2` (no directive) | `SESSION_ALIVE after 15.0s` |

Variant A reproduces today's behavior (session killed on markdown); variant B proves a long-running session survives a replica markdown — the behavior requested in #1516.

### DIFF
--- a/automation/molecule/tests/roles/confd/variables/default_server.yml
+++ b/automation/molecule/tests/roles/confd/variables/default_server.yml
@@ -1,0 +1,101 @@
+---
+# 🚀 Render the real confd haproxy.tmpl.j2 template and assert the default-server
+# directive reacts to haproxy_default_server.
+# 🎯 Guards issue #1516 for the confd-managed copy: on_marked_down="" must drop
+# "on-marked-down shutdown-sessions"; the default must keep it.
+
+# 📂 Ensure tmp directory exists
+- name: Molecule.tests.roles.confd.variables.default_server | Ensure tmp directory exists
+  run_once: true
+  delegate_to: localhost
+  ansible.builtin.file:
+    path: "./tmp"
+    state: directory
+
+# 🔧 Baseline scalar vars required to render the template
+# (confd renders server entries via Go templating inside {% raw %}, so no
+# postgres_cluster inventory is needed here.)
+- name: Molecule.tests.roles.confd.variables.default_server | Set baseline render vars
+  run_once: true
+  delegate_to: localhost
+  ansible.builtin.set_fact:
+    haproxy_maxconn: { global: 100000, master: 10000, replica: 10000 }
+    haproxy_timeout: { client: "60m", server: "60m" }
+    haproxy_listen_port: { master: 5000, replicas: 5001, replicas_sync: 5002, replicas_async: 5003, stats: 7000 }
+    haproxy_bind_address: 10.0.0.10
+    cluster_vip: ""
+    patroni_restapi_protocol: http
+    pgbouncer_install: false
+    pgbouncer_listen_port: 6432
+    patroni_maximum_lag_on_replica: "100MB"
+    confd_default_server_template: "{{ playbook_dir }}/../../roles/confd/templates/haproxy.tmpl.j2"
+
+- name: Molecule.tests.roles.confd.variables.default_server | Set haproxy_default_server (default)
+  run_once: true
+  delegate_to: localhost
+  ansible.builtin.set_fact:
+    haproxy_default_server:
+      inter: "3s"
+      fastinter: "1s"
+      fall: 3
+      rise_master: 4
+      rise_replica: 2
+      on_marked_down: "shutdown-sessions"
+
+- name: Molecule.tests.roles.confd.variables.default_server | Render default config
+  run_once: true
+  delegate_to: localhost
+  ansible.builtin.template:
+    src: "{{ confd_default_server_template }}"
+    dest: "./tmp/haproxy.tmpl.default_server.default.cfg"
+
+- name: Molecule.tests.roles.confd.variables.default_server | Read default-server lines (default)
+  run_once: true
+  delegate_to: localhost
+  ansible.builtin.shell: "grep 'default-server' ./tmp/haproxy.tmpl.default_server.default.cfg"
+  register: confd_ds_default
+  changed_when: false
+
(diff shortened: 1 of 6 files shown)


vitabaks:
As I understand it, the main motivation is to avoid interrupting requests running on a replica with replication lag.

In other scenarios, however, keeping the connection open after a replica failure may cause connections to remain stuck on the failed server. Have you tested this scenario?

pen-pal:
> As I understand it, the main motivation is to avoid interrupting requests running on a replica with replication lag.
> 
> In other scenarios, however, keeping the connection open after a replica failure may cause connections to remain stuck on the failed server. Have you tested this scenario?

Good point — that's a real tradeoff, not a free win, so I tested it explicitly against a live HAProxy 3.0 + backend, covering both a health-check markdown and actual backend failures (clean crash vs. silent partition).

## Results

| `default-server` | Failure mode | Active session outcome |
|------------------|--------------|------------------------|
| `... on-marked-down shutdown-sessions` (default) | silent failure (host down / partition, no RST) | closed at markdown (~5s) |
| `on_marked_down: ""` | silent failure | **held until `timeout server`** (default `60m`) |
| `on_marked_down: ""` | clean crash / restart (kernel RST) | closed in ~2s |
| `on_marked_down: ""` | transient lag (replica healthy at TCP, marked down by lag check) | session continues — the intended behavior |

Full run with raw output is attached: `haproxy_failover_tests.md`.

## Takeaways

- `on-marked-down shutdown-sessions` mainly protects against **silent** failures (hard power-off, network partition) where no RST ever reaches HAProxy. Without it, a session to such a node stays pinned until `timeout server` — which defaults to `60m` in this repo. That's exactly the "stuck on a failed server" case you raised.
- For a **clean** crash/restart the RST propagates and the session closes in ~2s regardless, so removing the directive is harmless there.
- The motivating case (transient replication lag: replica is healthy at the TCP layer, just marked down by the lag health check) keeps the session alive, which is the whole point.

## Why this is still safe to merge

- The default is unchanged: `on_marked_down: "shutdown-sessions"`. Existing deployments behave exactly as before - nobody is exposed unless they explicitly opt out.
- I'll add a doc note: if you set `on_marked_down: ""`, pair it with a shorter `haproxy_timeout.server` so a silently-failed replica can't pin a connection for the full 60m.

Issue #1359, opened 2025-11-25
[Feat] Console (UI): refresh interval


vitabaks:
### Problem

Currently, a manual **refresh** button is available, and the Console also supports automatic updates through the `_POLLING_INTERVAL` [settings](https://github.com/vitabaks/autobase/tree/master/console/ui#configuration).

<img width="2344" height="718" alt="Image" src="https://github.com/user-attachments/assets/ed77bbef-9893-4b0a-96bd-4341d2a6ff34" />

However, there is no option to change the refresh rate directly in the UI.

### Description

Add option to change the refresh rate directly in the UI (for example, the way it’s done in Grafana).

### Importance

nice to have

vitabaks:
Done https://github.com/autobase-tech/autobase/pull/1563

PR #1563, opened 2026-05-30
feat: UI refresh interval

pen-pal:
## Issue

Relates to #1359 

## Summary

Add a per-user refresh-interval dropdown to the console UI so users can choose how often dashboards re-poll — or turn polling off entirely. The choice is saved to `localStorage` and persists across sessions.

## What changes

- **New `RefreshGroup` MUI component** — a Grafana-style joined `[ ↻ Refresh | 1m ▾ ]` widget: a manual refresh button on the left, an interval dropdown on the right, sharing one rounded border with a vertical divider between segments.
- **`RefreshIntervalSelect`** — Button + Menu (instead of TextField/Select), value + chevron only. Options: `Off / 5s / 10s / 30s / 1m / 5m / 15m / 30m / 1h`. Includes a `normalizeInterval` helper that snaps env-supplied defaults to the nearest dropdown option so the Select never renders a blank/out-of-range value.
- **`pollingIntervalSlice`** (Redux) with `localStorage` persistence.
- **`useQueryPolling` hook**: reactive (reads the slice; switches interval on change) and honors `Off`.
- Wired into **4 page toolbars** — clusters list, cluster overview, operations list, operation log. The first three use `RefreshGroup`; operation log uses `RefreshIntervalSelect` standalone (no manual refresh on that page).
- **`vitest.config.ts`** scopes Vitest to `src/**/*.{test,spec}.{ts,tsx}` so Playwright e2e specs aren't picked up as unit tests.

## Caveats — UI refresh interval is not backend freshness

This dropdown controls **how often the browser re-fetches data from the API**, not how often the backend re-reads from the source.

- **Cluster views** — backend freshness is bounded by `PG_CONSOLE_CLUSTERWATCHER_RUNEVERY` (default `1m`).
- **Operation logs** — bounded by `PG_CONSOLE_LOGWATCHER_RUNEVERY` (default `1m`).

If those service watchers run every `1m`, setting the UI to `5s` or `10s` will mostly re-read the same DB state. Use the UI dropdown to opt into a faster cadence only when the backend watcher is also tuned down, or to turn polling off entirely. Documented in `console/ui/README.md` so future readers don't get the wrong impression of data timeliness.

## Tests

- Vitest default env switched to `jsdom`.
- Unit tests for `pollingIntervalSlice`.
- Playwright e2e suite covering the dropdown + persistence.

## How to use

Open any page with the toolbar dropdown, pick an interval, and it's applied across the app — including after reload.

## Screenshots

Screenshot 2026-05-31 at 12 01 15
Screenshot 2026-05-31 at 12 01 35
Screenshot 2026-05-31 at 12 02 08
Screenshot 2026-05-31 at 12 02 33


vitabaks:
@pen-pal maybe it's worth making it closer to the Grafana style?

Move the interval selection to the right and make it narrower in width. To make refresh and interval look like one button.

image

vitabaks:
**Findings**

- P1: `vitest.config.ts` does not exclude Playwright specs, and the new `e2e/refresh-interval.spec.ts` file matches Vitest’s default `*.spec.ts` pattern. As a result, `yarn vitest` picks up the e2e test as a unit test and fails. Fix by either limiting Vitest `include` to `src/**/*.test.ts(x)` or excluding `e2e/**`.

- P2: `pollingIntervalSlice.ts` accepts environment defaults as arbitrary numbers, but `RefreshIntervalSelect` renders only fixed `POLLING_INTERVAL_OPTIONS`. If a user sets something like `PG_CONSOLE_CLUSTERS_POLLING_INTERVAL=15000`, polling will run every 15s, but the select will receive a value with no matching `MenuItem`. MUI will show an out-of-range warning and the displayed value may be blank or incorrect. Normalize env defaults to allowed options or add the env value to the option list.

vitabaks:
@pen-pal  Note on refresh intervals: the new UI polling interval only controls how often the browser re-fetches data from the API. It does not make the backend data fresher.

For cluster-related views, freshness is still bounded by `PG_CONSOLE_CLUSTERWATCHER_RUNEVERY`. For operation logs, it is bounded by `PG_CONSOLE_LOGWATCHER_RUNEVERY`. If those service watchers run every `1m`, setting the UI refresh to `5s` or `10s` will mostly re-read the same DB state and may give users a misleading impression that data is updated that often.

pen-pal:
> @pen-pal Note on refresh intervals: the new UI polling interval only controls how often the browser re-fetches data from the API. It does not make the backend data fresher.
> 
> For cluster-related views, freshness is still bounded by `PG_CONSOLE_CLUSTERWATCHER_RUNEVERY`. For operation logs, it is bounded by `PG_CONSOLE_LOGWATCHER_RUNEVERY`. If those service watchers run every `1m`, setting the UI refresh to `5s` or `10s` will mostly re-read the same DB state and may give users a misleading impression that data is updated that often.

how do you think I should address it?
Any suggestions?

vitabaks:
> how do you think I should address it? Any suggestions?

So far, I’m not sure it’s worth extending the API to allow overriding the interval. It might be sufficient to document this behavior in the UI README instead.

pen-pal:
> > how do you think I should address it? Any suggestions?
> 
> So far, I’m not sure it’s worth extending the API to allow overriding the interval. It might be sufficient to document this behavior in the UI README instead.

made updates on Readme and ui as similar to grafana refresh button

vitabaks:
Moved the refresh button to the right

image

PR #1583, opened 2026-06-22
Move shared_buffers to postgresql_parameters

vitabaks:
Add cluster-wide PostgreSQL memory settings to the common defaults. Remove the corresponding host-level entries from local_postgresql_parameters (replace them with an empty list while keeping commented examples) to avoid duplicate or conflicting overrides.

By default, PostgreSQL parameters stored in DCS should be the single source of truth for cluster-wide configuration. Necessary for integration with the  UI platform.

If needed, you can still override settings for a specific host by defining the local_postgresql_parameters variable.


vitabaks:
FYI @pat-s

PR #1586, opened 2026-06-24
Ensure /etc/apt/sources.list.d exists before deb822_repository

### LINKED ISSUE #1584: [Bug] /etc/apt/sources.list.d/ does not exist
### Bug description

The add_repository role uses ansible.builtin.deb822_repository (introduced in commit 9473b4a) to configure APT repositories on Debian-based systems. This module writes .sources files to /etc/apt/sources.list.d/, but no task ensures this directory exists before the module call. On minimal systems or freshly provisioned nodes where this directory is absent, the module fails with FileNotFoundError during os.rename() of the temporary file to the destination.

ERROR:
FileNotFoundError: [Errno 2] No such file or directory:
b'/home/backoffice_ansible/.ansible/tmp/ansible-tmp-.../tmpp2h07vb_'
-> b'/etc/apt/sources.list.d/local-repo.sources'

### Expected behavior

The Add repository task should succeed regardless of whether /etc/apt/sources.list.d/ already exists. The directory should be created automatically by the role before the deb822_repository module is called.

### Steps to reproduce

1. Provision a Debian 12 (or Ubuntu 24.04+) node where /etc/apt/sources.list.d/ does not exist (e.g., a minimal Docker container or a freshly bootstrapped system without the apt package fully configured).
2. Define apt_repository in the inventory's group_vars/all.yml:
apt_repository:
  - repo: "deb https://apt.postgresql.org/pub/repos/apt/ {{ ansible_distribution_release }}-pgdg main"
    key: "https://apt.postgresql.org/pub/repos/apt/ACCC4CF8.asc"
3. Run the deploy_pgcluster playbook against this node.
4. Observe the failure: the Add repository task crashes with FileNotFoundError.

### Installation method

Command line

### System info

- Autobase version: 2.8.0 (all versions since the switch to deb822_repository in commit 9473b4a)
- Target OS: Debian 12 (Bookworm) / Ubuntu 24.04+
- Ansible module: ansible.builtin.deb822_repository

### Additional info

**Affected files**

All locations where deb822_repository is called without ensuring the parent directory exists:
1. automation/roles/add_repository/tasks/main.yml — primary Add repository task (Debian block, line 20)
2. automation/roles/add_repository/tasks/extensions.yml — Add TimescaleDB repository (line 7) and Add Citus repository (line 35)
3. automation/roles/pg_probackup/tasks/main.yml — Add pgdg repository (line 16)
4. automation/roles/consul/tasks/install_linux_repo.yml — Add hashicorp repository (line 81)

**Suggested fix**

Add a file task to ensure /etc/apt/sources.list.d/ exists before any deb822_repository call, for example in add_repository/tasks/main.yml inside the Debian block:
```
- name: Ensure /etc/apt/sources.list.d directory exists
  ansible.builtin.file:
    path: /etc/apt/sources.list.d
    state: directory
    mode: '0755'
```
The same task should be added to the other three files listed above, or alternatively a shared pre_tasks include could be used.

pen-pal:
## Issue
Fixes #1584

## Summary
On minimal Debian/Ubuntu hosts, repository setup fails with `FileNotFoundError` because `ansible.builtin.deb822_repository` writes `<name>.sources` files into `/etc/apt/sources.list.d/` but does not create that directory itself. The directory is normally provided by the `apt` package, so the failure only shows up on stripped-down images/containers where it is absent. This adds an idempotent directory-ensure task immediately before every `deb822_repository` call so provisioning succeeds on those systems.

## What changes
- New `ansible.builtin.file: { state: directory, mode: "0755" }` task that ensures `/etc/apt/sources.list.d` exists, inserted **before** each of the 5 `deb822_repository` invocations.
- Each new task is placed **inside the same block** as the repo task it guards, so it inherits the block's `become`, `tags`, and `when`; its task-level `when` is set to exactly match the adjacent `deb822_repository` task. This guarantees the directory is created in **every** code path that adds a repo — including tag-filtered runs (e.g. `--tags timescaledb`) and the arch-gated Citus path — and never runs when no repo is added.
- Wired into all 4 affected role files / 5 call sites:
  - `add_repository/tasks/main.yml` — `apt_repository` loop
  - `add_repository/tasks/extensions.yml` — TimescaleDB and Citus
  - `pg_probackup/tasks/main.yml` — pgdg/PostgresPro repo
  - `consul/tasks/install_linux_repo.yml` — HashiCorp repo (runs under `become: true`)

## Caveats
- Debian/Ubuntu only — `deb822_repository` is a Debian-family module; the RedHat/`yum_repository` paths are untouched.
- Idempotent and zero-churn on normal systems: the directory almost always already exists, in which case the task reports `ok` and changes nothing.
- No `owner`/`group` is set, so an existing directory's ownership is never altered; `mode: "0755"` is the conventional permission for that path.

## Tests
Validated end-to-end in a Debian 12 container (`geerlingguy/docker-debian12-ansible`, ansible-core 2.19.11) by reproducing the exact bug condition — base OS repos in legacy `/etc/apt/sources.list` and `/etc/apt/sources.list.d/` removed — and running the real `add_repository` tasks with the default pgdg `apt_repository`:

| Scenario | Code | `.d` dir | Result |
| --- | --- | --- | --- |
| Reproduce | pre-fix (`main`) | absent | **FAIL** — `FileNotFoundError ... -> /etc/apt/sources.list.d/apt-postgresql-org.sources` (matches the issue) |
| Fix | this PR | absent | **PASS** — dir created (`0755`), `.sources` written, `apt update` succeeds |
| No-regression | this PR | present | **PASS** — dir task reports `ok` (no churn), full run green |
| Idempotency | this PR | present (re-run) | **PASS** — `changed=0`, fully idempotent |

Also: `yamllint -c .config/.yamllint` passes on all four changed files; `mode: "0755"` quoting matches the repo convention.

## How to reproduce / verify
1. On a Debian 12 / Ubuntu 24.04+ host (or container) without `/etc/apt/sources.list.d/`, define an `apt_repository` entry and run the `deploy_pgcluster` playbook.
2. **Before:** the *Add repository* task fails with `FileNotFoundError: [Errno 2] No such file or directory: '.../local-repo.sources'`.
3. **After:** the directory is created first and the repository is added successfully.

### DIFF
--- a/automation/roles/add_repository/tasks/extensions.yml
+++ b/automation/roles/add_repository/tasks/extensions.yml
@@ -1,16 +1,19 @@
 ---
 # Extension Auto-Setup: repository
 
+# The deb822_repository module does not create /etc/apt/sources.list.d/;
+# ensure it exists once before adding any extension repository.
+- name: Ensure /etc/apt/sources.list.d directory exists
+  ansible.builtin.file:
+    path: /etc/apt/sources.list.d
+    state: directory
+    mode: "0755"
+  when: ansible_os_family == "Debian"
+  tags: add_repo, timescaledb, timescale, citus
+
 # TimescaleDB (if 'enable_timescale' is 'true')
 - block:
     # Debian based
-    - name: Ensure /etc/apt/sources.list.d directory exists
-      ansible.builtin.file:
-        path: /etc/apt/sources.list.d
-        state: directory
-        mode: "0755"
-      when: ansible_os_family == "Debian"
-
     - name: Add TimescaleDB repository
       ansible.builtin.deb822_repository:
         name: "timescaledb"
@@ -39,13 +42,6 @@
 # Citus (if 'enable_citus' is 'true')
 - block:
     # Debian based
-    - name: Ensure /etc/apt/sources.list.d directory exists
-      ansible.builtin.file:
-        path: /etc/apt/sources.list.d
-        state: directory
-        mode: "0755"
-      when: ansible_os_family == "Debian" and ansible_architecture in ["x86_64", "amd64"]
-
     - name: Add Citus repository
       ansible.builtin.deb822_repository:
         name: "citusdata"
--- a/automation/roles/add_repository/tasks/extensions.yml
+++ b/automation/roles/add_repository/tasks/extensions.yml
@@ -4,6 +4,13 @@
 # TimescaleDB (if 'enable_timescale' is 'true')
 - block:
     # Debian based
+    - name: Ensure /etc/apt/sources.list.d directory exists
+      ansible.builtin.file:
+        path: /etc/apt/sources.list.d
+        state: directory
+        mode: "0755"
+      when: ansible_os_family == "Debian"
+
     - name: Add TimescaleDB repository
       ansible.builtin.deb822_repository:
         name: "timescaledb"
@@ -32,6 +39,13 @@
 # Citus (if 'enable_citus' is 'true')
 - block:
     # Debian based
+    - name: Ensure /etc/apt/sources.list.d directory exists
+      ansible.builtin.file:
+        path: /etc/apt/sources.list.d
+        state: directory
+        mode: "0755"
+      when: ansible_os_family == "Debian" and ansible_architecture in ["x86_64", "amd64"]
+
     - name: Add Citus repository
       ansible.builtin.deb822_repository:
         name: "citusdata"
--- a/automation/roles/consul/tasks/install_linux_repo.yml
+++ b/automation/roles/consul/tasks/install_linux_repo.yml
@@ -77,6 +77,13 @@
         - ansible_os_family == "Debian"
         - "'python3-debian' not in ansible_facts.packages"
 
+    - name: Ensure /etc/apt/sources.list.d directory exists
+      ansible.builtin.file:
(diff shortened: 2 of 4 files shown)


automation/roles/add_repository/tasks/extensions.yml
```diff
@@ -4,6 +4,13 @@
 # TimescaleDB (if 'enable_timescale' is 'true')
 - block:
     # Debian based
+    - name: Ensure /etc/apt/sources.list.d directory exists
```
vitabaks: I think it's enough to move the task from the block once, so as not to duplicate the code.

automation/roles/consul/tasks/install_linux_repo.yml
```diff
@@ -77,6 +77,13 @@
         - ansible_os_family == "Debian"
         - "'python3-debian' not in ansible_facts.packages"
 
+    - name: Ensure /etc/apt/sources.list.d directory exists
+      ansible.builtin.file:
+        path: /etc/apt/sources.list.d
+        state: directory
+        mode: "0755"
+      when: "ansible_os_family|lower == 'debian'"
```
vitabaks:
```suggestion
      when: ansible_os_family == "Debian"
```

Issue #1587, opened 2026-06-25
[Bug] postgres user is locked by default when installed from repository (RHEL,Debian)


illmouse:
### Bug description

When `enable_ssh_key_based_authentication: true`, the `ssh_keys` role sets up SSH keys and authorized_keys for the `postgres` user, but SSH connections as `postgres` still fail with `Permission denied (publickey,keyboard-interactive)`.

The root cause is that the `postgres` OS user is created by the PostgreSQL package with a **locked password** (`!` in `/etc/shadow`). PAM's `pam_unix.so` account module rejects locked accounts before SSH public key verification is attempted, so the key in `authorized_keys` is never checked.

### Expected behavior

After the `ssh_keys` role completes, the `postgres` user should be able to SSH between cluster nodes using key-based authentication, which is required by Patroni for replication and cluster operations.

### Steps to reproduce

1. Deploy a new cluster on Debian 12 with `enable_ssh_key_based_authentication: true` (default)
2. Attempt SSH as postgres user between nodes:
   ```
   sudo -u postgres ssh postgres@<other-node> echo OK
   ```
3. Observe failure: `Permission denied (publickey,keyboard-interactive)`
4. Check syslog: `User postgres not allowed because account is locked`

### Installation method

Command line

### System info

- OS: Debian 12 (bookworm), also affects RHEL/AlmaLinux
- Autobase version: 2.8.0+
- PostgreSQL package: any version — the OS package always creates the `postgres` system user with a locked password

### Additional info

Relevant log entry from ansible job log:
```
postgres@192.168.1.100: Permission denied (publickey,keyboard-interactive).
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: unexplained error (code 255) at io.c(232) [sender=3.2.7]
```

The `cron` role has a similar fix for RHEL (PR #1360) that unlocks the `postgres` user for cron job execution, but it only applies to RedHat and only when cron jobs are defined. The SSH use case is not covered.

PR #1588, opened 2026-06-25
Fix/unlock postgres user for ssh (issue #1587)

illmouse:
### Summary

Fix: unlock postgres OS user account in ssh_keys role to enable SSH key-based authentication (closes issue #1587)

### Description

On Debian (and RHEL), the `postgres` system user is created by the PostgreSQL package with a locked password (`!` in `/etc/shadow`). While this prevents password-based login, PAM's `pam_unix.so` account module also blocks locked accounts before SSH public key verification is attempted.

This means that even though the `ssh_keys` role correctly generates SSH keys and populates `authorized_keys`, all SSH connections as `postgres` fail with `Permission denied` when `enable_ssh_key_based_authentication` is enabled.

#### Changes

Added two tasks to `roles/ssh_keys/tasks/main.yml`, after user creation and before SSH key generation:

1. **Check password lock state** — runs `getent shadow` to detect if the account is locked
2. **Unlock the locked password** — sets `password: '*'` (invalid hash, prevents password login) and `password_lock: false` (removes `!` from shadow)

Both tasks are guarded by:
- `ssh_key_user == 'postgres'` — only affects the postgres user
- `ssh_key_user_shadow.stdout is match('^postgres:!+')` — only runs when the account is actually locked

#### Why this approach

- Idempotent: only triggers if the account is locked
- Secure: `password: '*'` is an invalid hash, so password-based login is still impossible
- Consistent with existing fix in `cron` role (PR #1360), but scoped to OS-agnostic SSH setup
- Minimal: 10 lines of code, no new dependencies

#### Related

- PR #1360 — similar fix for cron role (RHEL-only)

### DIFF
--- a/automation/roles/ssh_keys/tasks/main.yml
+++ b/automation/roles/ssh_keys/tasks/main.yml
@@ -35,6 +35,21 @@
         state: present
         system: true
 
+    - name: Check password lock state for "{{ ssh_key_user | default('') }}" user
+      ansible.builtin.command: "getent shadow {{ ssh_key_user | default('') }}"
+      register: ssh_key_user_shadow
+      changed_when: false
+      no_log: true
+
+    - name: Unlock the locked password for "{{ ssh_key_user | default('') }}" user (required for SSH key auth)
+      ansible.builtin.user:
+        name: "{{ ssh_key_user | default('') }}"
+        password: '*'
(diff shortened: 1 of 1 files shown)


vitabaks:
Ok, thanks!

PR #1601, opened 2026-07-09
fix: enable TLS on an existing cluster without pre-flight health-check deadlock

fasiihullah:
## Problem

When enabling TLS on an **already-deployed, previously non-TLS cluster** (`tls_cert_generate: false` → `true`) via `config_pgcluster.yml`, the play aborts on the pre-flight health check with *"No healthy leader node detected"* — even though the cluster is perfectly healthy.

**Root cause (chicken-and-egg):**
- `patroni_restapi_protocol` is derived from `tls_cert_generate` (`common/defaults/main.yml`), so enabling TLS makes it resolve to `https`.
- But the **running** Patroni REST API is still serving `http` — it isn't reconfigured/restarted to `https` until later in the same run.
- The pre-flight `[Prepare] Get Patroni Cluster Leader Node` task probes `https://…:8008/leader`, hits an SSL error (`WRONG_VERSION_NUMBER`, status -1), and the `any_errors_fatal` `The Patroni cluster is unhealthy` gate aborts the play **before** it can switch Patroni to https.

Reproduction:
```
# existing cluster deployed with tls_cert_generate: false
# set tls_cert_generate: true  (+ tls_cert_regenerate: true for one run to generate certs)
ansible-playbook playbooks/config_pgcluster.yml -i <inventory>
# => aborts on '[Prepare] Get Patroni Cluster Leader Node' / 'The Patroni cluster is unhealthy'
```

## Fix

Make the pre-flight check protocol-aware of the **running** API instead of only the desired protocol:

1. A lightweight detection probe tries the configured protocol (`status_code: [200, 503]`, so both leader and replica count as reachable).
2. A `set_fact` derives `patroni_restapi_effective_protocol`, falling back to the opposite protocol if the first probe didn't reach the API.
3. The existing `Get Patroni Cluster Leader Node` check uses the effective protocol.

The downstream leader-status gate and primary/secondary grouping are **unchanged** — only the protocol used to reach the API is made adaptive. This keeps the blast radius minimal.


vitabaks:
@fasiihullah ok, this is suitable for `tls_cert_generate: false → true`. Have you tried the `tls_cert_generate: true → false` scenario?

fasiihullah:
@vitabaks I've only tested false → true so far.

vitabaks:
> (+ tls_cert_regenerate: true for one run to generate certs)

We will no longer need to set tls_cert_regenerate to enable TLS. Only if they need to be explicitly recreated.

Commit: https://github.com/autobase-tech/autobase/pull/1601/commits/75c4965d1fa64ff1cf3b3b154cb229620da9c63b

PR #1671, opened 2026-08-26
fix: require console auth for the embedded SQL editor (/dbdesk)

himanshu-ntml:
## Summary

The embedded SQL editor (dbdesk-studio) is reverse-proxied by the console UI's nginx under `/dbdesk/` **without any authentication**, while `/api/` requires the console token. Because dbdesk-studio's own API is reachable through that proxy, an unauthenticated client can do everything the editor can — including reading the stored cluster credentials:

```
$ curl -s http://<console>/dbdesk/api/connections
[{"id":"…","name":"db1","type":"postgres","options":{"host":"…","port":5000,"database":"db1","user":"postgres","password":"<cleartext>","sslMode":"require"}, …}]
$ curl -s -X POST http://<console>/dbdesk/api/connections/<id>/query -H 'content-type: application/json' -d '{"query":"select 1"}'
{"rows":[…]}   # arbitrary SQL as the cluster superuser
```

The connection is created automatically by the console (`console/service/internal/watcher/dbdesk.go` → `POST /api/connections` with the `postgres` password), so every deployed cluster is exposed this way, on both the all-in-one `autobase/console` image and the compose setup.

## Fix

nginx now guards `/dbdesk/` with `auth_request`, delegating to the API's existing token check (`GET /api/v1/version` → 200/401), so there is a single source of truth for what a valid token is. Since the editor is loaded by navigation and issues its own XHR, it cannot carry the `Authorization` header; the UI therefore mirrors the console token into an `autobase_dbdesk` cookie scoped to the editor path (`Path=/dbdesk/; SameSite=Strict`, `Secure` on https), set on login and for sessions restored from `localStorage`, and cleared on logout.

- `console/ui/nginx/nginx.conf`: internal `location = /_dbdesk_auth` + `auth_request` in `location ^~ /dbdesk/` (uses the existing `REPLACE_ME_WITH_API_HOST/PORT` placeholders, so `env.sh` needs no change).
- `console/ui/src/shared/lib/dbdeskAuthCookie.ts`: `setDbdeskAuthCookie` / `clearDbdeskAuthCookie` (path derived from `VITE_DBDESK_URL`).
- `pages/login`, `features/logout-button`, `app/router/PrivateRouterWrapper`: set / clear / refresh the cookie.

No new configuration, no API changes. `auth_request` is part of the stock nginx build used by the images.

## Verification

Applied the nginx change to a running `autobase/console:2.10.0` container (all-in-one) and checked:

| request | before | after |
|---|---|---|
[...]

PR #1672, opened 2026-08-26
console/ui: validate the token via API instead of embedding it in the build

himanshu-ntml:
## What & why

The console authorization token is substituted into the built JS at container start (`env.sh`: `VITE_AUTH_TOKEN` → `REPLACE_ME_WITH_AUTH_TOKEN` in every `*.js`), and login only compares the entered value to that embedded constant on the client. So the token is readable from a static asset **before** login:

```
$ curl -s http://<console>/assets/functions-<hash>.js | grep -o 'VITE_AUTH_TOKEN:`[^`]*`'
VITE_AUTH_TOKEN:`<the real token>`
```

I understand the console is intended for private/trusted networks, so this isn't framed as a remote exploit — but on a shared network it means the login screen doesn't actually gate anything (anyone who can reach the port can lift the token and call the API), and it puts the secret in cacheable static assets. Not shipping it costs nothing and makes the login meaningful.

## Change

Validate the entered token against the API instead of an embedded copy:

- **`pages/login`** — on submit, store the token and validate it by calling a bearer-gated endpoint (`GET /version`) via the existing `useLazyGetVersionQuery`; keep it only on success, clear it + show the existing `invalidToken` toast on 401. The token field is now `type="password"`.
- **`app/router/PrivateRouterWrapper`** — gate on *presence* of a stored token (which is only stored after that validation); real enforcement stays server-side, where every API call already carries the bearer and a bad one gets 401.
- **`shared/config/constants`** — drop the `AUTH_TOKEN` constant.
- **`env.sh`, `.env`, `.env.production`** — stop resolving/substituting `VITE_AUTH_TOKEN` into the build.
- **`ui/README`** — the `_FILE` convention note no longer lists `VITE_AUTH_TOKEN`.

`PG_CONSOLE_AUTHORIZATION_TOKEN` (and its `_FILE` form) is unchanged — only the browser copy is removed. No API or backend change.

### Why `GET /version`

The CE API is token-only — there's no dedicated `/auth/login` or `/auth/validate` route, and every endpoint sits behind the same bearer-token middleware (`/version` returns `401` without a valid token, `200` with one). `/version` is just the lightest already-gated GET, so hitting it validates the token server-side. (In the EE console the purpose-built route is `POST /auth/login`; an EE UI change would validate against that instead.)

## Verification

- `vite build` on this branch: `REPLACE_ME_WITH_AUTH_TOKEN` and `VITE_AUTH_TOKEN` are **absent** from `dist/`; other runtime placeholders (`REPLACE_ME_WITH_API_URL`, …) and the `/version` call are still present.
- Confirmed live against a running console that `GET /api/v1/version` returns **401** for a missing/incorrect token and **200** only for the correct one, so the validation genuinely rejects a bad token.
- `eslint` clean on the changed files. Prettier: the one remaining warning on `login/ui/index.tsx` is a pre-existing unwrapped `<Typography>` line already present on `main` (left untouched to keep the diff focused). `tsc --noEmit` emitted no errors.
- The Playwright e2e (`refresh-interval.spec.ts`) sets `localStorage.token` directly and bypasses the login page, so it is unaffected.

## Notes

- The EE UI (`console_ui_ee`) ships the same `env.sh` / `.env.production`, so it embeds the token the same way — this change applies there too (validating against `/auth/login`).
- Optional follow-up (not in this PR to keep it focused): a global 401 handler that clears the token and redirects to login, so a token that becomes invalid *after* login doesn't leave the user on a page that only 401s.
- Touches `pages/login` and `PrivateRouterWrapper`, which also appear in #1671 (dbdesk `/dbdesk/` auth). They're independent concerns; whichever merges second needs a trivial rebase — happy to do that.


vitabaks:
This is a really important fix, thanks!

PR #1674, opened 2026-08-28
pgbackrest: add pgbackrest_repo_external for restoring from an unmanaged external repo

fasiihullah:
## Summary
Adds pgbackrest_repo_external (default false) so pgbackrest_repo_host can point at an existing/external repository (e.g. a read-only repo) used purely as a restore source, without this collection managing it.

Setting pgbackrest_repo_host requires the host to be in the [pgbackrest] inventory group, which makes this collection manage that host (regenerate its pgbackrest.conf, run stanza-create, install cron). Setting pgbackrest_repo_external: true skips that requirement and skips managing the repo, while client nodes still get repo1-host/repo1-host-user injected into their local pgbackrest_conf.


fasiihullah:
Closing to resubmit with a more narrowly-scoped fix.

PR #1694, opened 2026-09-14
fix(upgrade): correct doubled log path in recovery_target LSN check

fasiihullah:
## Bug

In `automation/roles/upgrade/tasks/recovery_target.yml` (blue-green major-version upgrade), the "Check the PostgreSQL log file" task builds a doubled, broken path:

```
grep -F 'recovery stopping after WAL location (LSN) "{{ target_lsn }}"' {{ postgresql_log_dir }}/{{ pg_log_filename.stdout }}
```

The preceding task, "Get current PostgreSQL log file", already resolves `pg_log_filename.stdout` to a full absolute path:

```sql
select case
when pg_current_logfile() is null then ''
when pg_current_logfile() like '/%' then pg_current_logfile()
else current_setting('data_directory') || '/' || pg_current_logfile()
end
```

So prepending `{{ postgresql_log_dir }}/` on top of that produces a path like:

```
/var/log/postgresql//var/log/postgresql/postgresql-Mon.log
```

`grep` fails with `No such file or directory`, and the task always reports the recovery target as unreached — even when the actual PostgreSQL log confirms it was reached (`recovery stopping after WAL location (LSN) "..."` is present). This aborts the blue-green upgrade playbook at stage 4/8, every time, regardless of whether recovery actually succeeded.

## Fix

Remove the redundant `{{ postgresql_log_dir }}/` prefix — `pg_log_filename.stdout` is already a complete, absolute path.


vitabaks:
Exactly, I missed that. Thanks!

PR #1700, opened 2026-09-16
fix(aws): remove post-creation EC2 resource tagging

vitabaks:
This change removes the standalone `ec2_tag` tasks that updated EC2 instances and tagged attached EBS volumes, network interfaces, and Spot requests after resource creation.

For regular EC2 instances, `amazon.aws.ec2_instance` already applies ownership tags to the instance and its EBS volumes atomically as part of the `RunInstances` request. The subsequent tagging tasks were therefore redundant for these resources and required permissions to tag already existing AWS resources.

Removing these tasks allows `ec2:CreateTags` to be restricted to resource creation using the `ec2:CreateAction` condition. This prevents Autobase credentials from assigning ownership tags to unrelated resources after they have been created and then gaining lifecycle control over them.

ENIs will remain without ownership tags. Autobase does not manage network interfaces independently, and `delete_on_termination: true` ties each ENI lifecycle to its EC2 instance.

This also means changes to `cloud_provider_tags` are no longer propagated to existing EC2 resources after deployment. New instances and EBS volumes continue to receive the configured tags during creation.

Issue https://github.com/autobase-tech/autobase/issues/1699


vitabaks:
### Remaining limitations for strict tag-based IAM policies

This change removes redundant post-creation tagging for on-demand EC2 instances and EBS volumes. These resources already receive `managed-by: autobase` atomically through `RunInstances`.

Deployments using a strict IAM policy that requires:

```json
"aws:RequestTag/managed-by": "autobase"
```

should account for the following remaining limitations:

- **EC2 network interfaces:** `amazon.aws.ec2_instance` does not include `network-interface` in `RunInstances.TagSpecifications`. ENIs therefore remain untagged. Autobase does not manage ENIs independently, and `delete_on_termination: true` ties their lifecycle to the EC2 instance. IAM policies should exclude ENIs from ownership-tag-dependent operations.

- **Spot instances:** `amazon.aws.ec2_spot_instance` applies tags atomically only to the Spot request. The resulting instance is tagged later by the `AWS: Rename the EC2 Spot instance` task, while its EBS volumes and ENIs cannot receive ownership tags at creation. A policy that permits `CreateTags` only with `ec2:CreateAction` is therefore not compatible with the current Spot implementation. Full support requires migration to a `RunInstances`-based launch template flow.

- **NLB target groups:** AWS supports tags in `CreateTargetGroup`, but `community.aws.elb_target_group` currently creates the target group first and calls `AddTags` separately. Consequently, `elasticloadbalancing:CreateAction = CreateTargetGroup` cannot be used with this module. Until the module supports tags during creation, `AddTags` should be restricted by target-group ARN/name prefix, `aws:RequestTag/managed-by = autobase`, and `aws:TagKeys`.

- **S3 backup buckets:** AWS now supports tags in `CreateBucket`, but `amazon.aws.s3_bucket` currently applies bucket tags after creation. Strict tag-on-create policies therefore require either module support for `CreateBucket` tags or a narrowly scoped exception for the configured backup bucket.

`aws:RequestTag/managed-by = autobase` validates the value being requested, but by itself does not prevent a post-creation tagging permission from being used against an unrelated resource. Where post-creation tagging remains necessary, the permission must also be restricted to the exact resource ARN or the narrowest possible naming pattern.

PR #1703, opened 2026-09-17
fix(aws): migrate Spot from ec2_spot_instance to ec2_instance + insta…

### LINKED ISSUE #1699: [Bug] AWS: post-creation tagging allows resource hijacking via CreateTags
### Bug description

When creating EC2 instances, autobase tags EBS volumes, network interfaces, and Spot requests
in a dedicated `ec2_tag` task **after** the resources already exist - not during `RunInstances`.

Between the moment a resource is created and the moment it receives its `managed-by` tag,
any `CreateTags` permission scoped to "resources without an ownership tag" will also match
**every other untagged resource in the account**: EKS nodes, Karpenter instances, volumes
and load balancers created by other teams.

This means that whoever runs autobase can tag those foreign resources as `managed-by: autobase`
and immediately gain full lifecycle control over them - **terminate instances, modify volumes,
deregister targets** - because the ownership check in downstream operations only looks at tags,
not at who created the resource.

## Fix

**1. Tag volumes and ENIs at launch** (requires `amazon.aws >= 7.0.0`)

Add `tags` to each entry in `volumes[]` and `network_interfaces[]` inside the
`"AWS: Create EC2 instance"` task, then remove the standalone
`"AWS: Tag attached EBS volumes, network interfaces"` task.

`RunInstances` supports `TagSpecifications` for `instance`, `volume`, `network-interface`,
and `spot-instances-request` at launch time - tags and creation are atomic.

**2. Migrate Spot to `ec2_instance` + `instance_market_options: {market_type: spot}`**

The legacy `RequestSpotInstances` API used by `ec2_spot_instance` does not support
`TagSpecifications`, so Spot-launched volumes and ENIs cannot be tagged at creation
with the current implementation. The modern `RunInstances`-based path supports full
tagging at launch and is idempotent by instance name, removing the need for separate
wait and rename tasks.

**3. Fix NLB target group tagging**

`community.aws.elb_target_group` calls `AddTags` separately after `CreateTargetGroup`
and does not use the `Tags` parameter available in the `elbv2:CreateTargetGroup` API.
Fix: either patch the module to pass tags in `CreateTargetGroup`, or replace the module
call with a direct boto3 invocation.

### Expected behavior

All resources created by autobase (instances, volumes, network interfaces, Spot requests)
receive ownership tags at the moment of creation, as part of the same API call that creates them.

### Steps to reproduce

1.
2.
3.
...

### Installation method

Other

### System info

main branch

### Additional info

- AWS `RunInstances` - `TagSpecification.N` supports: `instance`, `volume`, `network-interface`, `spot-instances-request` https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html
- `elbv2:CreateTargetGroup` accepts `Tags` in the same call (boto3 `ElasticLoadBalancingv2.Client`) https://docs.aws.amazon.com/boto3/latest/reference/services/elbv2/client/create_target_group.html
- `RequestSpotInstances` API - marked legacy in AWS documentation https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestSpotInstances.html
- `community.aws.elb_target_group` source confirms `AddTags` is called after creation: https://github.com/ansible-collections/community.aws/blob/main/plugins/modules/elb_target_group.py

auroaj:
Replace three tasks (create spot request, wait, rename) with a single `ec2_instance` call using `instance_market_options: {market_type: spot}`.

`RunInstances` supports `TagSpecifications` for instance, volume, network-interface, and spot-instances-request at launch time, so Name and ownership tags are applied atomically — the post-creation rename step is no longer needed.

Also fixes `system_volume_size` being hardcoded to 100 in the legacy spot path - the underlying bug (ansible-collections/amazon.aws#1949) remains not fixed in `ec2_spot_instance`, but `ec2_instance` handles integer variables correctly.

Closes #1699


vitabaks:
@auroaj Thanks for the contribution.

Welcome to Autobase project!

vitabaks:
I couldn't find support for `instance_market_options` in `amazon.aws.ec2_instance` — it is not present in either the module documentation or the current source code. The documentation also explicitly states that `ec2_instance` does not support creating Spot instances directly.

Docs: https://docs.ansible.com/projects/ansible/latest/collections/amazon/aws/ec2_instance_module.html
Source: https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_instance.py

The available workaround is to configure `instance_market_options: {market_type: spot}` in an EC2 Launch Template and then pass that template to `ec2_instance` via `launch_template`.

This approach was also suggested by the `amazon.aws` maintainers here:
https://github.com/ansible-collections/amazon.aws/issues/1338#issuecomment-1381574256

Launch Template module:
https://docs.ansible.com/projects/ansible/latest/collections/amazon/aws/ec2_launch_template_module.html

auroaj:
It is described in AWS API doc.
So we again blocked with community ansible playbook.

vitabaks:
As I mentioned in my comment above, you can try the `ec2_launch_template` module for Spot instances.

